PR #5, opened 2021-02-08
initial support for remote daemons

dkackman:
This PR modifies the electron app such that it can connect to a daemon running somewhere other than `localhost`.

(moved from https://github.com/Chia-Network/chia-blockchain/pull/842)

Of note:
- the regular `config.yaml` file is used to locate the daemon
   - added the following to the `ui` section to find and secure the wss connection
````yaml
  daemon_host: localhost
  daemon_port: 55400
  daemon_ssl:
    private_crt: config/ssl/daemon/private_daemon.crt
    private_key: config/ssl/daemon/private_daemon.key
````
   - the remote daemon must be reachable at `wss://<daemon_host>:<daemon_port>`
- the UI needs a copy of the daemon's certs
  - the `ui` config section on the ui machine might look like
````yaml
  daemon_host: farmer
  daemon_port: 55400
  daemon_ssl:
    private_crt: config/ssl/ui/private_daemon.crt
    private_key: config/ssl/ui/private_daemon.key
````
- the config file is located based on the version placed in `package.json` by the build OR in the source tree during development. _Not sure this will work in all scenarios so that could use particular review_
- The code will not attempt to start or stop the daemon if it is not at `localhost`
- I didn't change the code that gets the cert paths from the locally hosted daemon via `stdout`. That code path can probably be removed and just rely on the config file (where the python gets it anyway) but I didn't want to perturb that for the time being

### CODE REVIEW

**Review by hoffmang9 — APPROVED** [review]
With the comment that I'd like remote.md moved to the wiki in a subsequent release, this looks good to me. Needs @Yostra or @seeden to look at it too.

File: src/util/config.js [hunk 1]
```diff
@@ -18,7 +18,7 @@ function loadConfig() {
 
     // package major will be 0 until release
     if (sv.major == 0) {
-      version = `beta-1.0b${sv.patch}`;
+      version = `beta-0.1.${sv.patch}`;
```
hoffmang9, line 21: This is an aberration I caused to get the Windows Installer build correctly. However we're probably moving to rc1 next.

dkackman: I figure I'll keep that little bit of code synced to the latest version until release. If for no other reason than I need it and if anyone else uses this fork they'll need it too

dkackman:
Cool. I'll move the howto over to wiki once this makes it into a release.

PR #14, opened 2021-02-11
Sub block to block

seeden:
fixed opening of the blocks
used correct block fields
fixed sync progress
changed timestamp format


hoffmang9 (approved):
Lgtm

PR #15, opened 2021-02-11
Ru translate

Bibop182:
- The message.po file has been translated from English into Russian (ru)
- GUI -> Wallets -> Receive Address:
        The width of the button "Copy" next to the public address bar is no longer fixed. Now the size automatically adjusts to the content text. Previously, when the button size was fixed (width: 70), the translation did not fit into it
- Adds the Russian language to the locale toggle dropdown list

### DIFF
--- a/.linguirc
+++ b/.linguirc
@@ -4,7 +4,7 @@
     "include": ["<rootDir>/src"],
     "exclude": ["**/node_modules/**"]
   }],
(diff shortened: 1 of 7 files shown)


seeden (approved):
Looks good. Thank you

Bibop182:
Many thanks 👍

PR #31, opened 2021-02-15
Ms.no sub blocks

mariano54:


mariano54:
I'm not sure if this is the correct workflow for PRs. Should I submit PRs to dev or main?

hoffmang9:
I think @seeden originally wanted to have main be the release but there seem to be issues with pinning to the branch. As such I think we should drop dev and have branches just go into main and we'll update the submodule in chia-blockchain dev.

Thoughts?

hoffmang9:
Thinking about how we have this wrong in main in chia-blockchain, I'm going to retarget this to dev and merge it there.

PR #51, opened 2021-02-17
Small fixes3

seeden:
used bech32m for puzzle_hash
added link to chiaexplorer for puzzle_hash
used last height farmed only from reward transactions


hoffmang9 (approved):
It threw me that bech32m was in bech32 but now that I get that, this looks good.

PR #71, opened 2021-02-21
Fix a tiny problem in a tooltip text

jespino:
This text was generating problem for some translators because makes no sense. I changed the text with what I think was the real intention.


hoffmang9 (approved):
Thanks!

PR #84, opened 2021-02-24
Test audit and locales

hoffmang9:
I'm sure you have a better way but I'm very afraid of `npm run locale:extract` changing the scm version in the installers and I've seen some misbehavior there. The alternative of the GUI in an install having that not run is also a bad idea.

Let's go ahead and squash merge this for the short run and then you can tell me what you'd like to do instead.


seeden (approved):
looks good

PR #94, opened 2021-02-25
Increase start timeout when talking to daemon

mariano54:


hoffmang9 (approved):
ok but we should see if we can shorten that down some

PR #106, opened 2021-02-26
Fix locale check - windows is "broken"

hoffmang9:


hoffmang9:
This now correctly checks if the locales are compiled. It found that Window is not and that matches with that I saw in the RC3/RC4 installer builds.

I seem to recall you mentioning a way we can automate this @seeden ?

seeden:
@hoffmang9  can I close it? We are doing extraction during the build now

hoffmang9:
Yeah. I wanted to capture that check though and now have.

PR #112, opened 2021-03-02
Not open multiple close dialogs.

### LINKED ISSUE #93: Double dialog when closing the UI on linux
The dialog that asks "are you sure you want to close" appears twice if I click close twice

jespino:
Fixes #93


hoffmang9 (approved):
+1

PR #115, opened 2021-03-02
updated full node connections and add peer

seeden:


hoffmang9:
image
Enter an address is a little too small for an IPv6 address like:
2600:1f14:8f8:6500:1461:8f02:97ed:c474

hoffmang9 (approved):
Looks good once LGTM.com and the dialog box are handled.

PR #117, opened 2021-03-03
UI search block

seeden:
fixed issue multiple alerts when block does not exists
moved search input to block card


hoffmang9 (approved):
Looks good

PR #157, opened 2021-03-14
check venv

dkackman:
At start-up this adds a check to the electron app to make sure it is either packaged (i.e. installed on mac or windows) or running in `venv`

https://trello.com/c/cbUbmZ0r/1022-can-npm-run-electron-check-for-the-virtualenv-environment-variable-and-fail-with-must-be-in-venv

_not tested on MAC but since it uses 'guessPackaged` should work like the daemon search does_

### DIFF
--- a/src/electron-starter.js
+++ b/src/electron-starter.js
@@ -39,7 +39,16 @@ if (!setupEvents.handleSquirrelEvent()) {
     }
   };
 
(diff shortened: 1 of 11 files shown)


hoffmang9 (approved):
Thank you so much on this one Don!

hoffmang9:
I'm going to test this on mac too and sneak it into rc7

PR #168, opened 2021-03-15
Removed extract locales

seeden:
extract locales before `npm run build` and `npm run dev`
removed locales javascript files from the git because they will be generated automatically
added locales js files into the gitignore


hoffmang9 (approved):
+1

PR #181, opened 2021-03-17
Update messages.po

Dravenex:
I hope you like it. ;)


seeden:
Hi @Dravenex.
First of all, thank you for your work.
I am sorry but we are not accepting translations via GitHub PR.
If you want to improve translations you need to add your translations here https://crowdin.com/project/chia-blockchain.
The reason is that we have automatic PR from Crowdin and we are not able to do it in the opposite way.

hoffmang9:
You should be able to copy and paste over at crowdin. We need to update the README - sorry about that.

Issue #184, opened 2021-03-19
[BUG] Opening dev tools causes and exception


dkackman:
![image](https://user-images.githubusercontent.com/5160233/111851256-8b891200-88e0-11eb-89d5-568a66cd08f2.png)

dkackman:
mainWindow isn't in scope in this block after I changed the quick exit checks. fix incoming
````
          {
            label: i18n._(/*i18n*/{id: "Developer"}),
            submenu: [
              {
                label: i18n._(/*i18n*/{id: "Developer Tools"}),
                accelerator:
                  process.platform === "darwin"
                    ? "Alt+Command+I"
                    : "Ctrl+Shift+I",
                click: () => mainWindow.toggleDevTools()
              }
            ]
          },
````

seeden:
already merged

PR #185, opened 2021-03-19
fix Chia-Network/chia-blockchain-gui#184

dkackman:
Fixes the bug where the open devtools menu will error on open because mainWindow is not in scope.


seeden (approved):
looks good

seeden:
thanks :)

PR #200, opened 2021-03-24
First crack at build scripts.

richardkiss:


richardkiss:
This doesn't have a chance of working until we have an `sdist` of `chia-blockchain` in pypi that works. Once we have that, we can see how this goes (and then we can get rid of the gui build in the `chia-blockchain` repo).

richardkiss:
There's a decent chance that this will work again, which would allow us to remove the gui submodule & gui build scripts from `chia-blockchain`.

richardkiss:
This is more of a proof of concept at this point... probably want to compare against build scripts in `chia-blockchain`, since the version these scripts are based upon are 20 days old. But if you want me to work on this some more, let me know.

PR #203, opened 2021-03-24
used compatible version for optional chaining

seeden:
partial fox for issue (https://github.com/Chia-Network/chia-blockchain/issues/1384) with starting electron app on Ubuntu

### DIFF
--- a/src/util/config.js
+++ b/src/util/config.js
@@ -2,6 +2,7 @@ const yaml = require('js-yaml');
 const fs = require('fs');
 const os = require('os');
 const path = require('path');
+const lodash = require('lodash');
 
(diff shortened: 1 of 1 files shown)


hoffmang9 (approved):
Looks good. What's the underlying problem on Ubuntu?

seeden:
Ubuntu has preinstalled the old version of Node which has no support for JavaScript optional chaining. We are transpiling the application. But the main starting script for the Electron application is not using it. I will change it in the next refactoring cycle.

You can find more information about this issue here https://github.com/Chia-Network/chia-blockchain/issues/1384#issuecomment-805916769.

PR #207, opened 2021-03-26
used env variable for local farming instead of changing source code

seeden:
used env variable for local farming instead of changing source code


hoffmang9 (approved):
Took me a sec to grok the use case but now makes total sense

hoffmang9:
Markdown needs lint changes - https://github.com/Chia-Network/chia-blockchain-gui/pull/207/checks?check_run_id=2203976498#step:4:267

PR #221, opened 2021-03-30
used locales instead of two_letters_code

seeden:


hoffmang9 (approved):
That should fix it!

seeden:
Yeah. I need to wait for new PR from Crowdin service as well and update configuration file in the GUI

PR #230, opened 2021-04-05
Src -> Chia

mariano54:
Directory was renamed


hoffmang9 (approved):
Approved assuming @richardkiss doesn't have to roll back/sideways to deal with pyinstaller.

PR #238, opened 2021-04-07
starting of electron application

seeden:
replaced and  refactored how we are starting electron application
added support for typescript
removed overcomplicated waiting for react build
used webpack instead of react-scripts
used latest npm packages and fixed problem with outdated packages


hoffmang9 (approved):
+1

Issue #293, opened 2021-04-21
[GUI] Privacy shield for last height farmed


ddegirmenci:
Hi! Many people are posting screenshots of their GUI on reddit and elsewhere, and this can only keep increasing as more people get their first few XCH. Since the "last height farmed" field is a potential privacy vulnerability to link a wallet to an online identity, I feel a privacy shield could be a quick win here. 

Maybe something like "show last height farmed only on mouseover"?

A tooltip to point out the risk can also be helpful.

PR #346, opened 2021-05-04
Allow farm cards to be toggled to hide values

sayomaki:
As mentioned in #293, I think having a toggle-able button to change the state of the display to show or hide the value would be a good idea.

This is done with a new property on the `FarmCard` component, and this is reusable for other cards in the farm section if there are any other privacy concerns from users who are posting their screenshots of their farm page.


seeden:
This will be part of the pool release.

Issue #72, opened 2021-02-21
Need UI hints or tooltips for basic info


mariano54:
Users are confused about these three points:
- Transactions are not available within the first 10k blocks (it will be more in mainnet). Maybe gray out the "Create Transaction" pane?
- What is the transaction fee to use? (it can be set to 0 or something really low like 0.001
- How long do transactions take? (Maybe add this in the confirmation message or in a tooltip)

seeden:
@mariano54 what is a good description for "How long do transactions take?"

hoffmang9:
It can take 55 to 65 seconds for a next transaction. Maybe just say "On average there is one minute between each transaction block. Unless there is congestion you can expect your transaction to be included in less than a minute."

etr2460:
It looks like this hasn't been addressed yet, I'm going to try taking some of it on since it's labeled as "good first issue"

seeden:
PR already merged. I am closing

PR #360, opened 2021-05-07
Add transaction time tooltip

etr2460:
Adds a tooltip to the Create Transaction card with details about how long the transaction should take.

Unfortunately I wasn't able to get a test environment working locally; running `npm run dev` resulted in the dev server crashing. Do I need to also be running other chia-blockchain repos to test locally? Advice would be much appreciated, thanks!

See https://github.com/Chia-Network/chia-blockchain-gui/issues/72

to: @hoffmang9 @seeden


seeden (approved):
Looks good. Thanks

PR #433, opened 2021-05-25
This PR fixes the problem of adding a button to save the drawing file path selection style error when using Chinese

Hoinor:
Fix the problem that the style of adding the button to save the plots file path selection is wrong when using Chinese

Since the style error only occurs in the two files, I used the inline style

After fix https://imgtu.com/i/gzLKVx
Before fix https://imgtu.com/i/gzLrRg


seeden:
I will change it to nowrap because min-width can be buggy when user change font size or use zoom

PR #442, opened 2021-05-29
remove find config by version logic

dkackman:
Remove the logic that still tried to find the config folder based on `chia version`, a holdover from pre-release versions. Now it just looks for 'mainnet' yet still respects `CHIA_ROOT`.

### DIFF
--- a/src/electron/main.tsx
+++ b/src/electron/main.tsx
@@ -110,7 +110,7 @@ if (!handleSquirrelEvent()) {
   // if any of these checks return false, don't do any other initialization since the app is quitting
   if (ensureSingleInstance() && ensureCorrectEnvironment()) {
     // this needs to happen early in startup so all processes share the same global config
-    chiaConfig.loadConfig(chiaEnvironment.getChiaVersion());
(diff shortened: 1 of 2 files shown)


hoffmang9 (approved):
Lgtm

src/electron/main.tsx
```diff
@@ -110,7 +110,7 @@ if (!handleSquirrelEvent()) {
   // if any of these checks return false, don't do any other initialization since the app is quitting
   if (ensureSingleInstance() && ensureCorrectEnvironment()) {
     // this needs to happen early in startup so all processes share the same global config
-    chiaConfig.loadConfig(chiaEnvironment.getChiaVersion());
+    chiaConfig.loadConfig('mainnet');
```
seeden, line 113: how it will work with testnet?

dkackman: CHIA_ROOT will have to be set, which is the case for this bit of code already.

seeden:
Thank you @dkackman for your PR

PR #444, opened 2021-05-30
updated deps

seeden:
fixed browsers list
removed unused styleguide
updated deps and fixed info related to vulnerabilities


hoffmang9 (approved):
lgtm

PR #461, opened 2021-06-12
fix missing isWindows import

gforceg:
fixes a bug:
when one or more plot exists, the plots screen is unusable (blank) on linux


gforceg:
# Before:
image

# After:
image

seeden:
Thank you @gforceg but I am going to close it.
We are already fixed deleting of the plots on the backend side and therefore we are going to fix it in another way via https://github.com/Chia-Network/chia-blockchain-gui/pull/463

PR #462, opened 2021-06-13
Update macOS icon

opayen:
I got annoyed by the fact that the Chia app icon on macOS doesn't follow the [macOS 11 Guildelines from Apple](https://developer.apple.com/design/human-interface-guidelines/macos/icons-and-images/app-icon/), so I updated it myself.

Before:
before

After:
after


hoffmang9:
Thanks!

opayen:
I see that some CI checks failed, is there any specific thing I should do?

seeden:
Thank you @opayen :)

PR #470, opened 2021-06-29
Pools

seeden:
welcome pools
added tons of updates


seeden:
I am not able to fix the npm audit because there is no new version. It is just related to the building process. Problematic packages are not used in production.

hoffmang9 (approved):
I gave it a cursory review - will check out and run soon too.

PR #477, opened 2021-07-06
Pools update

seeden:
added view payout instructions
added faucet information
added view pool login link
added points success (percent)
updated locales
fixed bugs


hoffmang9 (approved):
lgtm

PR #481, opened 2021-07-07
Pools update2

seeden:
fixed amount inputs
fixed list of plots (user new rpc)
added node id to list of plots
added plot refresh to hero page
simplified loading


hoffmang9 (approved):
AOK

PR #493, opened 2021-07-13
Show payout address for a PlotNFT in the Pool section

AlexSSD7:
Adds this:

image

**Why is it required?**
At Flexpool we use Payout Address as the farm identifier, and there is no way to easily gather it in GUI. This PR fixes that.

As a consideration for the future, it would be nice to have an ability to change the payout address without the need of editing `~/.config/chia/mainnet/config/config.yml` file.


aqk (commented):
Looks good. @seeden can you review it?

seeden (approved):

PR #508, opened 2021-07-27
Add toggle to hide NFTs without plots

davidsonmizael:
Adding option to hide NFTs contracts without any plots in the Pools tab.


emlowe:
Thanks for the PR - this is a good idea. I tried running this on my ubuntu box and I received an error about "Rendered more hooks than during the previous render" when I click on the "eye" icon. I have a single PlotNFT with zero plots.
 
We plan to implement a slightly different idea for hiding PlotNFTs

PR #512, opened 2021-08-01
Update messages.po

nontoxice:


hoffmang9:
Thank you for the pull request. We use Crowdin to manage the translations. Can you move these there and we'll incorporate them? https://crowdin.com/project/chia-blockchain

PR #540, opened 2021-09-11
Fixed issue where if 2 of the same words were generated 

jack60612:
If 2 of the same were generated then it would give an error and display the wrong words or possibly more.
This affects all keys that were generated through the gui


jack60612:
https://github.com/Chia-Network/chia-blockchain/issues/8439

hoffmang9 (approved):
Looks good and thanks!

wjblanke (approved):
aok

PR #550, opened 2021-09-26
cleaned up unfinished passphrase feature

lipalong:
Removed an unfinished work-in-progress passphrase feature that had gotten merged in too early.


wjblanke (approved):
aok

PR #558, opened 2021-10-11
temp fix for cc wallet for testing usage

jcteng:
wallets is a array not directory indexed by wallet_id use filter function to pick it .


seeden:
thank you.
we are going to handle cc wallets in a different way. therefore I am closing this PR

PR #570, opened 2021-10-29
Updating localized strings and package*.json with latest changes in main

paninaro:


package.json
```diff
@@ -168,5 +168,5 @@
             "last 1 safari version"
         ]
     },
-    "version": "1.2.6-dev67"
-}
+    "version": "1.2.11-dev20"
```
paninaro, line 171: I'm not sure where this change comes from or if we want to take it. Looks like it also removed the trailing newline at the end of the file.

src/locales/be-BY/messages.po
```diff
@@ -3050,10 +3050,6 @@ msgstr "сінхр."
 msgid "syncing"
 msgstr "ідзе сінхр."
 
-#: src/components/plot/add/PlotAddChoosePlotter.tsx:89
-msgid "thirdparty.plotter.disclaimer"
-msgstr ""
-
```
paninaro, line 3056: This string was inadvertently added when `chiaplotters_gui` was merged. I did a batch find/replace to remove them and then re-ran `npm run locale:extract && npm run locale:compile`

package.json
emlowe: it's best to remove the version string from package.json.

paninaro: Restored

emlowe (approved):

PR #571, opened 2021-11-02
Fix first launch hang with undefined daemon_rpc_ws when config.yaml hasn't been written yet

paninaro:
When launching for the first time (~/.chia/standalone_wallet not present), the GUI hangs due to an exception throw at Client.ts:57 wherein url is undefined.

This change exposes the default value for daemon_rpc_ws when the config isn't available yet.


emlowe (approved):
Looks ok in my testing.

seeden:
Fixed in a different way and already merged.

PR #573, opened 2021-11-05
Plotnft fees

seeden:
added support for plotnft fees 
fixed layout size

### DIFF
--- a/src/modules/plotNFT.ts
+++ b/src/modules/plotNFT.ts
@@ -120,11 +120,11 @@ export function pwAbsorbRewards(walletId: number, fee?: string) {
   };
 }
 
-export function pwSelfPool(walletId: number) {
+export function pwSelfPool(walletId: number, fee?: number) {
   return async (dispatch) => {
     const { data } = await async_api(
       dispatch,
-      pwSelfPoolMessage(walletId),
+      pwSelfPoolMessage(walletId, fee),
       false,
     );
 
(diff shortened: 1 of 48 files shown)


src/modules/plotNFT.ts
```diff
@@ -120,11 +120,11 @@ export function pwAbsorbRewards(walletId: number, fee?: string) {
   };
 }
 
-export function pwSelfPool(walletId: number) {
+export function pwSelfPool(walletId: number, fee?: number) {
   return async (dispatch) => {
     const { data } = await async_api(
       dispatch,
-      pwSelfPoolMessage(walletId),
+      pwSelfPoolMessage(walletId, fee),
```
paninaro, line 127: Since `fee?` is an optional, is there any issue with `pwSelfPoolMessage()` calling `format_message()` with a potentially undefined value?

src/modules/plotNFT.ts
```diff
@@ -148,6 +149,7 @@ export function pwJoinPool(
         poolUrl,
         relativeLockHeight,
         targetPuzzlehash,
+        fee,
```
paninaro, line 152: Same question regarding `fee?` being an optional and being passed into a function that may(?) not handle it being undefined.

paninaro (approved):
Left a couple of questions regarding `fee` possibly being undefined and whether that would cause any issues. If there's no concern there, everything looks good to me.

src/modules/plotNFT.ts
seeden: Yeah, I will add default value as well

paninaro (approved):

PR #575, opened 2021-11-12
Set webpack hash algorithm

emlowe:
Sets the webpack hash algorithm to SHA256 to avoid issues with node 17 and ERR_OSSL_EVP_UNSUPPORTED errors from openssl


seeden (approved):
LGTM

PR #576, opened 2021-11-15
IDEA: wallet integration with yubiko

everycrypto:
For almost 10 years I have been using yubiko key with lastpass to safeguard my passwords.
I suggest integrating access to the wallet (insert the 24 words seeds) with the above procedure.
The user starts the chia wallet and encounters a new item in the initial menu, the 2FA one with yubiko / lastpass.
In this way we would have the total portability of the wallet on any PC.
The only addition is the updated sqlite wallet / db, which the user must carry with him (32gb usb key) ... but it is a simple data transfer.


seeden:
Hi @everycrypto.
Please create a github issue with this idea. PR is for different purpose

PR #583, opened 2021-11-23
Fix Farm LastAttemptedProof - Changed sort sign from "-" to "+"

FilipeThePinho:
The last attempted proof list is updating the rows from bottom. Should be the reverse order, like the others. Please double check the change if it works.

Thank you


ChiaMineJP:
The records of last attempted proof are saved at `packags/api/src/services/Farmer.ts`.
I think you don't need to sort every time the component gets rendered.
I'll close this PR as it is too old to be merged, but if you're still interested in this issue, please re-open this issue.
I'll add more advice.

PR #584, opened 2021-11-23
Set electron-osx-sign to v0.6.0

emlowe:
Rather than refer to github for electron-osx-sign - set to v0.6.0 which was recently published.


wjblanke (approved):
aok

PR #585, opened 2021-11-24
 "xmldom": "github:xmldom/xmldom#0.7.0" sha256 changed

jcteng:
"xmldom": "github:xmldom/xmldom#0.7.0" sha256 changed  ,  update to         "xmldom": "github:xmldom/xmldom#0.7.5" to avoid error on npm install


seeden:
We already using new version in main

PR #627, opened 2022-01-27
Add offerpool.io to offer share dialog, fix share bug

mikehw:
This PR adds offerpool as an option in the offer share dialog.

Fixes a bug where sharing was broken when the dialog launched after creating an offer because the offer property was not yet set.


paninaro (approved):
Tested, and things look good to me. I have a small change I'd like to make (adding a button to go to Offerpool.io) but I can address that after merging.

Issue #633, opened 2022-02-22
Offer - Cancel on blockchain fee warning (beta1.3-dev284)


loppefaaret:
Chia beta 1.3 (dev284)
When the mempool is filled, and users want to cancel an offer on blockchain, they are warned by using a 0.00005 XCH fee:
![Screen Shot 2022-02-22 at 7 17 22 AM](https://user-images.githubusercontent.com/32275244/155176651-ecff9e4f-4de8-46ef-8958-60f97a0a5b91.png)

Whether there should be a warning at that level, a different message, or moving the warning threshold im not sure. But it does seem a bit out of place in its current form.

seeden:
I changed the threshold value to 100,000,000,000 mojos until we add fee estimator

PR #637, opened 2022-03-01
Small updates

seeden:
remove all rows from list of transactions #10360
fixed date MM/DD/YYYY  format in the list of blocks #10223
content updates #10155
fixed errors when trying to switch a PlotNFT from one self pooling to pool #10356
fixed open external in about dialogue
fixed lint
increased fee threshold warning (beta1.3-dev284) #633

### DIFF
--- a/packages/api-react/src/services/wallet.ts
+++ b/packages/api-react/src/services/wallet.ts
@@ -175,17 +175,17 @@ export const walletApi = apiWithTag.injectEndpoints({
       walletId: number;
       poolUrl: string;
       relativeLockHeight: number;
-      targetPuzzlehash?: string;
+      targetPuzzleHash?: string;
       fee?: string;
     }>({
-      query: ({ walletId, poolUrl, relativeLockHeight, targetPuzzlehash, fee }) => ({
+      query: ({ walletId, poolUrl, relativeLockHeight, targetPuzzleHash, fee }) => ({
         command: 'pwJoinPool',
         service: Wallet,
         args: [
           walletId,
           poolUrl,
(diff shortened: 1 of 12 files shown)


packages/api-react/src/services/wallet.ts
```diff
@@ -175,17 +175,17 @@ export const walletApi = apiWithTag.injectEndpoints({
       walletId: number;
       poolUrl: string;
       relativeLockHeight: number;
-      targetPuzzlehash?: string;
+      targetPuzzleHash?: string;
       fee?: string;
     }>({
-      query: ({ walletId, poolUrl, relativeLockHeight, targetPuzzlehash, fee }) => ({
```
paninaro, line 181: The `pw_join_pool` RPC looks like it expects the parameter name to be `target_puzzlehash`.  `pwJoinPool()` in Wallet.ts seems correct. Is this name change correct if the RPC expects `target_puzzlehash` as the param name?

seeden: Pool services return target_puzzle_hash which is converted to targetPuzzleHash but chia API is using target_puzzlehash which is targetPuzzlehash. I renamed it because it was not working

paninaro: Ahh ok, so pwJoinPool still uses `target_puzzlehash`, which is populated by the value in `targetPuzzleHash`. Sounds good.

seeden: yes :)

paninaro (approved):

PR #644, opened 2022-03-07
autocomplete tabbing fix

lipalong:
Made it so that on the import from mnemonics page, while a word in the dropdown list is highlighted, pressing tab autofills that word in the text field and the cursor moves to the next text field.


paninaro:
This is great @lipalong. I tested the change and it works as expected.

Is it accurate to state that only the changes in `Autocomplete.tsx` are relevant to this change? Can we undo all of the changes to `package.json/package-lock.json` and the `messages.js/po` files so that this PR is clean? If those files need to be updated, I'd prefer a separate PR for just those changes.

ChiaMineJP:
chia-blockchain-gui uses `lerna` and links multiple packages/workspace with it.
So deleting root `package.json` will stop everything from working.

PR #647, opened 2022-03-07
Invalidate txns+count when using "Delete Unconfirmed Transactions"

paninaro:
When using the "Delete Unconfirmed Transactions" action, pending transactions will continue to appear in the transaction list.

This fix will cause the useDeleteUnconfirmedTransactionsMutation hook to invalidate the Transactions cache, as well as the (newly added) TransactionCount cache.

Issue #648

### DIFF
--- a/packages/api-react/src/services/wallet.ts
+++ b/packages/api-react/src/services/wallet.ts
@@ -4,7 +4,7 @@ import onCacheEntryAddedInvalidate from '../utils/onCacheEntryAddedInvalidate';
 import normalizePoolState from '../utils/normalizePoolState';
 import api, { baseQuery } from '../api';
 
-const apiWithTag = api.enhanceEndpoints({addTagTypes: ['Keys', 'Wallets', 'WalletBalance', 'Address', 'Transactions', 'WalletConnections', 'LoggedInFingerprint', 'PoolWalletStatus', 'NFTs', 'OfferTradeRecord', 'OfferCounts']});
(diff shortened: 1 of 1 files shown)


packages/api-react/src/services/wallet.ts
```diff
@@ -630,6 +634,7 @@ export const walletApi = apiWithTag.injectEndpoints({
         args: [walletId],
       }),
       transformResponse: (response: any) => response?.count,
+      providesTags: ['TransactionCount'],
```
seeden, line 637: we should add walletId as id to providesTags id: walletId

packages/api-react/src/services/wallet.ts
```diff
@@ -222,16 +222,20 @@ export const walletApi = apiWithTag.injectEndpoints({
...
       walletId: number;
     }>({
       query: ({ walletId }) => ({
         command: 'deleteUnconfirmedTransactions',
         service: Wallet,
         args: [walletId],
       }),
+      invalidatesTags: [
+        { type: 'Transactions', id: 'LIST' },
+        'TransactionCount',
```
seeden, line 235: we should add wallet Id as id and invalidate TransactionCount just for specific Wallet.
it will be connected via providesTags where you need to add walletId as well {type: 'TransactionCount', id: walletId }

seeden (changes requested):
Please add walletId for invalidating transactionsCount for the specific wallet. it will safe invalidate for another wallets. more info in comments

paninaro:
> Please add walletId for invalidating transactionsCount for the specific wallet. it will safe invalidate for another wallets. more info in comments

Good call. I've updated the PR to associate the cached value with its corresponding walletId.

PR #652, opened 2022-03-09
autocomplete fix

lipalong:
Made it so that on the import from mnemonics page, while a word in the dropdown list is highlighted, pressing tab autofills that word in the text field and the cursor moves to the next text field.


seeden (approved):
LGTM

PR #655, opened 2022-03-14
fixed bech32 decode for payout address

seeden:
fixed bech32 decode for payout address
used one source for this functionality from @chia/core


paninaro (approved):
lgtm

PR #657, opened 2022-03-14
added missing event get_connections for farmer service

seeden:
added missing event get_connections for farmer service
fix for: https://github.com/Chia-Network/chia-blockchain/issues/10622


paninaro (approved):
lgtm

PR #661, opened 2022-03-24
Fix OfferSummary render failure when an offer isn't requesting anything

paninaro:


packages/wallets/src/components/offers/OfferSummary.tsx
```diff
@@ -44,10 +44,13 @@ export default function OfferSummary(props: Props) {
   const makerFee: number = summary.fees;
   const makerAssetInfo = makerEntries.length === 1 ? lookupByAssetId(makerEntries[0][0]) : undefined;
   const takerAssetInfo = takerEntries.length === 1 ? lookupByAssetId(takerEntries[0][0]) : undefined;
-  const makerAmount = makerEntries[0][0].toLowerCase() === 'xch' ? Number(mojoToChia(makerEntries[0][1])) : Number(mojoToCAT(makerEntries[0][1]));
-  const takerAmount = takerEntries[0][0].toLowerCase() === 'xch' ? Number(mojoToChia(takerEntries[0][1])) : Number(mojoToCAT(takerEntries[0][1]));
-  const makerExchangeRate = makerAssetInfo && takerAssetInfo ? takerAmount / makerAmount : undefined;
-  const takerExchangeRate = makerAssetInfo && takerAssetInfo ? makerAmount / takerAmount : undefined;
+  const makerAmount = makerEntries.length > 0 ? (makerEntries[0][0].toLowerCase() === 'xch' ? Number(mojoToChia(makerEntries[0][1])) : Number(mojoToCAT(makerEntries[0][1]))) : undefined;
```
ChiaMineJP, line 47: It seems `toLowerCase()` is not necessary here according to:
https://github.com/Chia-Network/chia-blockchain/blob/main/chia/wallet/trading/offer.py#L169
But it is good to make assurance.

paninaro: Yes, I think it's generally a good idea to ensure the data is in a form that you expect. This code could move around over time and might be applied to some data that isn't lowercased.

seeden, line 47: you can remove Number() because mojoToChia will always return number

seeden (changes requested):

PR #670, opened 2022-03-28
Added checkbox for keeping service running background after GUI closes

ChiaMineJP:
# Change summary
- Added a checkbox to toggle on/off whether to keep services running background
- Save checkbox value to prefs.yaml
- Restore checkbox value from prefs.yaml

# Alert
Users may easily forget that they have Chia services running in the background, even when they're trying to update the app. Perhaps it would be beneficial to notify people to ensure they shut down the Chia-blockchain app.

### DIFF
--- a/packages/gui/src/electron/main.tsx
+++ b/packages/gui/src/electron/main.tsx
@@ -478,16 +478,26 @@ if (ensureSingleInstance() && ensureCorrectEnvironment()) {
     // if (!guessPackaged()) {
     //   mainWindow.webContents.openDevTools();
     // }
-    mainWindow.on('close', (e) => {
+    mainWindow.on('close', async (e) => {
       // if the daemon isn't local we aren't going to try to start/stop it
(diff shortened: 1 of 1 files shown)


packages/gui/src/electron/main.tsx
```diff
@@ -496,22 +506,36 @@ if (ensureSingleInstance() && ensureCorrectEnvironment()) {
                 id: 'Are you sure you want to quit?',
               }
             ),
+            checkboxChecked: keepBackgroundRunning ?? false,
+            checkboxLabel: i18n._(/* i18n */ { id: 'Keep service running background' }),
```
paninaro, line 510:
```suggestion
            checkboxLabel: i18n._(/* i18n */ { id: 'Keep service running in the background' }),
```

paninaro (approved):
Tested, works great! Thanks @ChiaMineJP!

PR #673, opened 2022-04-02
added paste mnemonic option on import wallet from mnemonic page

lipalong:
added an option where users can paste their 24 mnemonic words to populate the input fields

### DIFF
--- a/packages/wallets/src/components/PasteMnemonic.tsx
+++ b/packages/wallets/src/components/PasteMnemonic.tsx
@@ -1,22 +1,13 @@
-import React, { useEffect } from 'react';
+import React from 'react';
 import {
-  Box,
   Button,
-  Checkbox,
   Dialog,
   DialogActions,
   DialogContent,
-  DialogContentText,
   DialogTitle,
-  FormControlLabel,
   TextField,
-  Tooltip,
 } from '@material-ui/core';
-import {
-  Help as HelpIcon,
-} from '@material-ui/icons';
-import { t, Trans } from '@lingui/macro';
-import { AlertDialog, useOpenDialog } from '@chia/core';
+import { Trans } from '@lingui/macro';
 
 type Props = {
   onSuccess: (mnemonicList: string) => void;
@@ -25,7 +16,6 @@ type Props = {
 
 export default function MnemonicPaste(props: Props) {
   const { onSuccess, onCancel } = props;
-  const openDialog = useOpenDialog();
   let mnemonicListInput: HTMLInputElement | null;
 
   async function handleSubmit() {
--- a/packages/wallets/src/components/PasteMnemonic.tsx
+++ b/packages/wallets/src/components/PasteMnemonic.tsx
@@ -6,7 +6,7 @@ import {
   DialogContent,
   DialogTitle,
   TextField,
-} from '@material-ui/core';
+} from '@mui/material';
 import { Trans } from '@lingui/macro';
 
 type Props = {
--- a/packages/wallets/src/components/PasteMnemonic.tsx
+++ b/packages/wallets/src/components/PasteMnemonic.tsx
@@ -1,0 +1,101 @@
+import React, { useEffect } from 'react';
+import {
+  Box,
+  Button,
+  Checkbox,
+  Dialog,
+  DialogActions,
+  DialogContent,
+  DialogContentText,
+  DialogTitle,
+  FormControlLabel,
+  TextField,
+  Tooltip,
+} from '@material-ui/core';
+import {
+  Help as HelpIcon,
+} from '@material-ui/icons';
+import { t, Trans } from '@lingui/macro';
+import { AlertDialog, useOpenDialog } from '@chia/core';
+
+type Props = {
+  onSuccess: (mnemonicList: string) => void;
+  onCancel: () => void;
+};
+
+export default function MnemonicPaste(props: Props) {
+  const { onSuccess, onCancel } = props;
+  const openDialog = useOpenDialog();
+  let mnemonicListInput: HTMLInputElement | null;
+
+  async function handleSubmit() {
+    const mnemonicList: string = mnemonicListInput?.value ?? "";
+    onSuccess(mnemonicList);
+  }
+
+  async function handleCancel() {
+    onCancel();
+  }
+
+  async function handleKeyDown(e: React.KeyboardEvent) {
+    const keyHandlerMapping: { [key: string]: () => Promise<void> } = {
+      'Enter' : handleSubmit,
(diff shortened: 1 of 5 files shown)


packages/wallets/src/components/PasteMnemonic.tsx
```diff
@@ -0,0 +1,101 @@
...
+import { AlertDialog, useOpenDialog } from '@chia/core';
+
+type Props = {
+  onSuccess: (mnemonicList: string) => void;
+  onCancel: () => void;
+};
+
+export default function MnemonicPaste(props: Props) {
+  const { onSuccess, onCancel } = props;
+  const openDialog = useOpenDialog();
```
ChiaMineJP, line 28: `openDialog` is not used.

packages/wallets/src/components/WalletImport.tsx
```diff
@@ -9,8 +9,26 @@ import {
...
+  DialogContentText,
+  DialogTitle,
+  Form,
+  Flex,
+  Logo,
+  TextField,
+  useOpenDialog,
+  useShowError,
+  useTrans
+} from '@chia/core';
```
ChiaMineJP, line 29: The following modules are not used:  
`Dialog`, `DialogActions`, `DialogContent`, `DialogContentText`, `DialogTitle`, `TextField`, `useShowError`.

packages/wallets/src/components/WalletImport.tsx
```diff
@@ -43,11 +63,47 @@ export default function WalletImport() {
...
+  const { fields, replace } = useFieldArray({
     control: methods.control,
     name: 'mnemonic',
   });
 
+  const submitMnemonicPaste = (mnemonicList: string) => {
+    closeMnemonicPaste();
+
+    let mList = mnemonicList.match(/\b(\w+)\b/g);
+    const intersection = mList.filter(element => options.includes(element));
```
ChiaMineJP, line 75: The type of `mList` is either an array or `null`.
For example, if you paste nothing and click import button on the dialog, `mnemonicList` will be `""`(empty string).
In that case, `mList` will be `null` because it will fail to `match` regexp and get an exception on the next line. (Trying to call `filter` on `null`)

ChiaMineJP:
Current issue:
https://user-images.githubusercontent.com/84098616/162012198-943852bd-0bd4-4dcb-82a6-b828465c469b.mov

This issue will be resolved once you modifies your code as I suggested in the above review.

packages/wallets/src/components/WalletImport.tsx
```diff
@@ -43,11 +56,60 @@ export default function WalletImport() {
...
+            Your pasted list does not include 24 valid mnemonic words.
+          </Trans>
+        </AlertDialog>
+      );
+    } else {
+      const mnemonic = intersection.map((word) => ({ word }));
+      replace(mnemonic);
+    }
+
+    // above abandon abstract absurd account acid alert alley blast bless blind blanket boost book busy burden burger census chair chalk dinner direct dirt discover
```
paninaro, line 94: Remove?

paninaro (approved):

packages/wallets/src/components/WalletImport.tsx
lipalong: ha, i'd taken it out once before but it snuck back in

paninaro (approved):

packages/wallets/src/components/WalletImport.tsx
```diff
@@ -43,11 +56,58 @@ export default function WalletImport() {
...
+          <Trans>
+            Your pasted list does not include 24 valid mnemonic words.
+          </Trans>
+        </AlertDialog>
+      );
+    } else {
+      const mnemonic = intersection.map((word) => ({ word }));
+      replace(mnemonic);
+    }
+  };
```
ChiaMineJP, line 93: The lines around `openDialog` seems reduntant and you can reduce the loc.
How about
```
const intersection = mList?.filter(element => options.includes(element));
if(!intersection || intersection.length !== 24){
  openDialog(...);
  return;
}
const mnemonic = intersectino.map((word) => ({ word }));
replace(mnemonic);
```

ChiaMineJP, line 93: Just a suggestion.
The lines around `openDialog` seems reduntant and you can reduce the loc.
How about
```
const intersection = mList?.filter(element => options.includes(element));
if(!intersection || intersection.length !== 24){
  openDialog(...);
  return;
}
const mnemonic = intersectino.map((word) => ({ word }));
replace(mnemonic);
```

packages/wallets/src/components/WalletImport.tsx
```diff
@@ -43,11 +56,58 @@ export default function WalletImport() {
...
   });
 
-  const { fields } = useFieldArray({
+  const { fields, replace } = useFieldArray({
     control: methods.control,
     name: 'mnemonic',
   });
 
+  const submitMnemonicPaste = (mnemonicList: string) => {
+    closeMnemonicPaste();
```
ChiaMineJP, line 65: Putting `closeMnemonicPaste()` here means regardless of whether a valid mnemonic is pasted or not, dialog will be closed either.
I think dialog should remain opened if the pasted mnemonic is invalid for it is easy to re-paste again instead of re-open 'paste-mnemonic' dialog over and over.

Could you move this `closeMnemonicPaste()` after `replace(mnemonic)` ?
If you do so, it will work like this:

https://user-images.githubusercontent.com/84098616/162473238-5aa92476-007b-491d-8d81-42eb672850b3.mov

ChiaMineJP, line 65: Putting `closeMnemonicPaste()` here means regardless of whether a valid mnemonic is pasted or not, dialog will be closed either.
I think dialog should remain opened if the pasted mnemonic is invalid for it is easy to re-paste again instead of re-open 'paste-mnemonic' dialog over and over.

Could you move this `closeMnemonicPaste()` after `replace(mnemonic)` ?
If you do so, it will work like this:

https://user-images.githubusercontent.com/84098616/162473238-5aa92476-007b-491d-8d81-42eb672850b3.mov

ChiaMineJP (changes requested):

ChiaMineJP:
@lipalong If you allow me, I'll push the changes that I described in the latest review to your branch.

lipalong:
> @lipalong If you allow me, I'll push the changes that I described in the latest review to your branch.

@ChiaMineJP Sure, feel free to push those changes. Thanks.

ChiaMineJP (approved):

seeden (approved):

PR #677, opened 2022-04-07
added paste offer file with hotkeys functionality to View an Offer page

lipalong:
added the ability to use cmd+v on mac or ctrl+v on linux/windows to paste an offer file directly

### DIFF
--- a/packages/gui/package-lock.json
+++ b/packages/gui/package-lock.json
@@ -41,6 +41,7 @@
                 "react-dom": "17.0.2",
                 "react-dropzone": "11.5.1",
                 "react-hook-form": "7.24.1",
+                "react-hotkeys-hook": "3.4.7",
                 "react-number-format": "4.9.1",
                 "react-redux": "7.2.6",
                 "react-router": "6.2.1",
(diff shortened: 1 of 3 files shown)


packages/wallets/src/components/offers/OfferImport.tsx
```diff
@@ -7,6 +7,7 @@ import { useGetOfferSummaryMutation } from '@chia/api-react';
 import { type OfferSummaryRecord } from '@chia/api';
 import OfferDataEntryDialog from './OfferDataEntryDialog';
 import fs, { Stats } from 'fs';
+import { useHotkeys } from 'react-hotkeys-hook';
```
paninaro, line 10: Is this dependency already present in main? If not, I think running `npx lerna add react-hotkeys-hook --scope @chia/gui` and then adding the package.json/package-lock.json files should do it.

lipalong: good catch, i forgot to include those files in the commit

paninaro (approved):

PR #683, opened 2022-04-13
fixed cat list

seeden:
fixed cat list
fixed heading size for sub cards

### DIFF
--- a/packages/wallets/src/hooks/useWalletsList.ts
+++ b/packages/wallets/src/hooks/useWalletsList.ts
@@ -1,5 +1,5 @@
 import { useMemo } from 'react';
-import { WalletType } from '@chia/api';
+import { WalletType, type Wallet } from '@chia/api';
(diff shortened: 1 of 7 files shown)


packages/wallets/src/hooks/useWalletsList.ts
```diff
@@ -95,10 +109,21 @@ export default function useWalletsList(search?: string): {
       return undefined;
     }
 
-    const baseWallets = wallets.filter((wallet) => ![WalletType.CAT, WalletType.POOLING_WALLET].includes(wallet.type));
+    const baseWallets = wallets.filter((wallet: Wallet) => ![WalletType.CAT, WalletType.POOLING_WALLET].includes(wallet.type));
```
paninaro, line 112: No change needed now, but just to let you know that there will also be DID and NFT wallets to consider here.

paninaro (approved):
lgtm

PR #693, opened 2022-04-18
first working error boundary

seeden:


packages/core/src/components/ErrorBoundary/ErrorBoundary.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+  render() {
+    if (this.state.hasError) {
+      const { stacktrace, error } = this.state;
+      const issueLink = `https://github.com/Chia-Network/chia-blockchain/issues/new?${qs.stringify({
+        labels: 'bug,GUI',
+        template: 'bug_report.yaml',
+        title: `[BUG] ${error.message}`,
+        ui: 'GUI',
+        logs: `${error.message}\n\n${stacktrace}`,
```
paninaro, line 66: Would it be possible to include window.location or something to indicate where the exception originated (beyond what's in the stacktrace, which isn't always useful). Maybe a `useLocation()` hook would be useful?

paninaro:
This is great @seeden! Thanks for adding this.

packages/core/src/components/ErrorBoundary/ErrorBoundary.tsx
seeden: I added current URL.

paninaro (approved):

paninaro (approved):

PR #705, opened 2022-04-26
Fix issue entering a maker fee in the offer editor.

paninaro:
The <Fee> component doesn't behave well if the bound data has an undefined default value. The form data now expects a string and sets the default value to an empty string.


seeden (approved):
LGTM

PR #716, opened 2022-05-04
Wallentx/electron builder gui

wallentx:


paninaro:
Electron builder work has been completed

Issue #719, opened 2022-05-04
[BUG] Cannot read properties of undefined (reading 'map')


mattn08:
### What happened?

Loaded GUI pointing at a remote farmer. Happens every time on-load after I pick my account.

### Version

1.3.5-dev102

### What platform are you using?

Linux

### What ui mode are you using?

GUI

### Relevant log output or stacktrace

```shell
Cannot read properties of undefined (reading 'map')

URL
#/dashboard/wallets/1

Stacktrace
at webpack://@chia/wallets/dist/esm.js:3800:46 _toConsumableArray(nonAddedStrayCats.map(function
at webpack://@chia/gui/node_modules/react-dom/cjs/react-dom.production.min.js:168:376 $h
at webpack://@chia/gui/node_modules/react/cjs/react.production.min.js:23:112 useLayoutEffect
at webpack://@chia/wallets/dist/esm.js:3747:13 useMemo(function
at webpack://@chia/wallets/dist/esm.js:4294:39 WalletsManageTokens
at webpack://@chia/gui/node_modules/react-dom/cjs/react-dom.production.min.js:157:136 Ch
at webpack://@chia/gui/node_modules/react-dom/cjs/react-dom.production.min.js:180:153 oi
at webpack://@chia/gui/node_modules/react-dom/cjs/react-dom.production.min.js:269:342 a=b.pendingProps;f=e._init;e=f(e._payload);b.type=e;f=b.tag=hk(e);a=lg(e,a);switch
at webpack://@chia/gui/node_modules/react-dom/cjs/react-dom.production.min.js:250:346 Pj
at webpack://@chia/gui/node_modules/react-dom/cjs/react-dom.production.min.js:250:277 Pj
```

seeden:
Fixed via https://github.com/Chia-Network/chia-blockchain-gui/pull/721

PR #721, opened 2022-05-05
added support for non defined stray cats

seeden:
[BUG] Cannot read properties of undefined (reading 'map') #719


paninaro (approved):
lgtm

PR #747, opened 2022-05-24
save and restore window state

dkackman:
This change uses the [electron-window-state](https://www.npmjs.com/package/electron-window-state) package to save and restore window state (size, position, maximized) in the `createWindow`. In the `close` lambda, it `unmanage`s the window so it doesn't save the smaller window size and position that gets set as the local node is being shut down.

The window state is saved in the root of the config folder `window-state.json`.


paninaro:
@dkackman Thanks for this enhancement! We have a planned work item to move the electron app's userData directory under CHIA_ROOT, so saving `window-state.json` into the config directory won't be necessary -- it can just be written somewhere within the userData directory.

dkackman:
> @dkackman Thanks for this enhancement! We have a planned work item to move the electron app's userData directory under CHIA_ROOT, so saving `window-state.json` into the config directory won't be necessary -- it can just be written somewhere within the userData directory.

Cool.  I'll adjust this and drop a comment when it's ready for another review.

dkackman:
@paninaro I reverted the last commit that did the CHIA_ROOT stuff. By default `electron-window-state` saves to userData, so the window state file should follow that around now.

paninaro:
Tested and is working well for me.

PR #757, opened 2022-05-31
Profile and passphrase settings

lipalong:
Added Profile renaming. Truncated Profile ids. Added visibility toggles to passphrase prompts. Added caps lock detection to passphrase prompts.

### DIFF
--- a/packages/gui/src/components/app/AppKeyringMigrator.tsx
+++ b/packages/gui/src/components/app/AppKeyringMigrator.tsx
@@ -163,12 +163,14 @@ export default function AppKeyringMigrator() {
             type={showPassphraseText1 ? "text" : "password"}
             InputProps={{
               endAdornment: (
-                <InputAdornment position="end">
-                  {showCapsLock && <div><KeyboardCapslockIcon /></div>}
-                  <IconButton onClick={() => setShowPassphraseText1(s => !s)}>
-                    <VisibilityIcon />
-                  </IconButton>
-                </InputAdornment>
+                <Flex alignItems="center">
+                  <InputAdornment position="end">
+                    {showCapsLock && <Flex><KeyboardCapslockIcon /></Flex>}
+                    <IconButton onClick={() => setShowPassphraseText1(s => !s)}>
+                      <VisibilityIcon />
+                    </IconButton>
+                  </InputAdornment>
+                </Flex>
               )
             }}
             fullWidth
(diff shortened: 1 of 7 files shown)


packages/gui/src/components/app/AppKeyringMigrator.tsx
```diff
@@ -114,12 +121,26 @@ export default function AppKeyringMigrator() {
     );
   }
 
+  function handleKeyDown(e: KeyboardEvent): void {
+    if (e.getModifierState("CapsLock")) {
+      setShowCapsLock(true);
+    }
+  }
+
+  const handleKeyUp = (event) => {
```
paninaro, line 130: Suggestion for consistency with the `handleKeyDown()` function

```suggestion
  function handleKeyUp(e: KeyboardEvent): void {
```

packages/gui/src/components/app/AppPassPrompt.tsx
```diff
@@ -65,6 +73,16 @@ export default function AppPassPrompt(props: Props) {
     if (e.key === 'Enter') {
       handleSubmit();
     }
+
+    if (e.getModifierState("CapsLock")) {
+      setShowCapsLock(true);
+    }
+  }
+
+  const handleKeyUp = (event) => {
```
paninaro, line 82: Suggestion for consistency with `handleKeyDown()`:

```suggestion
  function handleKeyUp(event: KeyboardEvent): void {
```
```

**Comment by paninaro** [comment]
Few comments re: the passphrase dialog enhancements. 

1. Ideally the passphrase changes would be in a separate PR from the DID fixes/enhancements. That would give us some flexibility when deciding whether one set of changes is ready to go vs having both of them ready together. Maybe for next time.
2. We should consider creating a reusable component (or multiple components) to avoid duplicating code. We can look into this in a subsequent PR.
3. The visibility button should reflect the current state by toggling between two icons. Looking at the options for MUI icons, I'd suggest `Visibility` (which is what's being used now), and `VisibilityOff`. We can address this separately also.

File: packages/gui/src/components/settings/ProfileView.tsx [hunk 3]
```diff
@@ -23,6 +24,49 @@ const StyledCard = styled(Card)(({ theme }) => `
...
+  const handleBlur = (event) => {
+    if (event.target.value.trim() === "") {
+      setEditedText(text);
+    } else {
+      setDid({ walletId: walletId, name: event.target.value});
+    }
+  }
+
+  return (
+    <input
```
paninaro, line 52: The styling looks out of place using a plain input element. Does TextField work?

image

paninaro (approved):
Approving so that we can get this merged. We can address the other fix-ups later.

PR #760, opened 2022-06-01
Add Ability to manually input pool contract address for plot creation

jack60612:


packages/gui/src/components/plot/add/PlotAddNumberOfPlots.tsx
```diff
@@ -309,6 +309,17 @@ export default function PlotAddNumberOfPlots(props: Props) {
               />
             </FormControl>
           </Grid>
+          <Grid xs={12} item>
+            <FormControl variant="filled" fullWidth>
+              <TextField
+                name="plotNFTContractAddr"
+                type="text"
+                variant="filled"
+                placeholder="Plot NFT Plot Target Address, This will be used if a PlotNFT is not selected."
```
paninaro, line 318: This is too long for a placeholder string. I would change it to just
`Plot NFT Plot Target Address`, or add a tooltip with an explanation next to the TextField. I'm ignoring the punctuation/capitalization typos since we should be using a shorter string.

paninaro (changes requested):
String change please.

packages/gui/src/components/plot/add/PlotAddNumberOfPlots.tsx
jack60612: Fixed

jack60612:
fix merge issues

packages/gui/src/components/plot/add/PlotAddNumberOfPlots.tsx
```diff
@@ -309,6 +309,17 @@ export default function PlotAddNumberOfPlots(props: Props) {
               />
             </FormControl>
           </Grid>
+          <Grid xs={12} item>
+            <FormControl variant="filled" fullWidth>
+              <TextField
+                name="plotNFTContractAddr"
+                type="text"
+                variant="filled"
+                placeholder="Plot NFT Plot Target Address"
```
paninaro, line 318: This should use the `t` macro for localization. You'll need to add `t` as an import from @lingui/macro, i.e.:

`import { Trans, t } from '@lingui/macro';`

```suggestion
                placeholder={t`Plot NFT Plot Target Address`}
```

jack60612: Fixed, thanks!

paninaro (approved):
lgtm

PR #792, opened 2022-06-16
added connection status buttons to wallet header

lipalong:
Added buttons to the wallet header which display popover windows for the list of full node connections and the list of wallet connections.


paninaro (approved):
Looks good. Tested locally after making a few changes.

PR #858, opened 2022-07-27
Set userData path within CHIA_ROOT. Support basic migration of leveldb data from old location.

paninaro:
The change updates the userData path used by the app to reside with the CHIA_ROOT directory, e.g. `~/.chia/mainnet/gui/Chia Blockchain`

An attempt is made to copy over leveldb data from the default userData location to the updated location, therein preserving user preferences like the farming/wallet mode selection, light/dark mode selection, etc.


seeden (approved):
LGTM

PR #876, opened 2022-08-04
added data-testid attributes

seeden:
added missing data-testid attributes for better testing


paninaro (approved):
lgtm, though I have concerns about the maintenance of keeping data-testid attributes up-to-date and unique, as well as ensuring that any new code includes these attributes.

seeden:
> lgtm, though I have concerns about the maintenance of keeping data-testid attributes up-to-date and unique, as well as ensuring that any new code includes these attributes.

Therefore id has always component name as prefix: ComponentName-action => Settings-tab-general

PR #881, opened 2022-08-07
added splitting NFT description on newlines.

Jsewill:
As an NFT creator, I found it necessary for my collection, and thought it advantageous for others, to preserve some formatting in the NFT description. Newlines are a simple, bare minimum, and easy to implement.

I'm not certain whether or not this is the best way to do this, as I rarely work with react/jsx. Feel free to keep the good and toss the bad, or roll your own.

### DIFF
--- a/packages/gui/src/components/nfts/detail/NFTDetail.tsx
+++ b/packages/gui/src/components/nfts/detail/NFTDetail.tsx
@@ -73,8 +73,8 @@ export default function NFTDetail() {
                   <Trans>Description</Trans>
                 </Typography>
 
-                <Typography>
-                  {metadata?.description?.split('\n').map((line,i) => <p key={i}>{line}</p>) ?? <Trans>Not Available</Trans>}
+                <Typography sx={{ whiteSpace : 'pre-line'}}>
+                  {metadata?.description ?? <Trans>Not Available</Trans>}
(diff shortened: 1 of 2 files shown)


packages/gui/src/components/nfts/detail/NFTDetail.tsx
```diff
@@ -74,7 +74,7 @@ export default function NFTDetail() {
                 </Typography>
 
                 <Typography>
-                  {metadata?.description ?? <Trans>Not Available</Trans>}
+                  {metadata?.description?.split('\n').map((line,i) => <p key={i}>{line}</p>) ?? <Trans>Not Available</Trans>}
```
seeden, line 77: Probably better way is too use pure CSS
sx={{ whiteSpace: 'pre-line'}}

seeden (changes requested):

packages/gui/src/components/nfts/detail/NFTDetail.tsx
Jsewill: I'm not sure how I missed that. Thanks for pointing that out! I've updated it.

paninaro (approved):
Tested. Working well for me. Thanks for the contribution @Jsewill!

PR #887, opened 2022-08-08
NFT <--> CAT offer support (#884)

paninaro:
* WIP

* NFT <--> CAT offer editor support

* NFT <--> CAT offer viewer support

* Default to NFTForToken when creating offer from an NFT


seeden (approved):
lgtm

PR #888, opened 2022-08-09
Show NFT wallets that are associated with a missing DID

paninaro:
Previously the NFT selection dropdown assumed that NFT wallets with an associated DID implied that the DID was owned by the user. This change handles the case where NFT wallets have associated DIDs that are missing (not owned by the user).


seeden (approved):
lgtm

PR #882, opened 2022-08-07
Properly truncate the NFT title if it's too long to display

paninaro:
With the fix:

NFT Gallery:
image

NFT Offer:
image

PR #890, opened 2022-08-09
Properly truncate the NFT title if it's too long to display (#882)

paninaro:
* Properly truncate the NFT title if it's too long to display

* Simplification recommended by seeden


seeden (approved):
lgtm

PR #894, opened 2022-08-10
Header sync state

seeden:
fixed wallet and fullnode state in the header
fixed https://github.com/Chia-Network/chia-blockchain-gui/issues/880


paninaro (approved):
Tested. Fixes sync/refresh issues I was seeing.

PR #895, opened 2022-08-10
Header sync state (#894)

paninaro:
* used correct wallet state

* fixed fullnode state

* added pollingInterval: 10000 for cats and list of wallets


seeden (approved):
Lgtm

PR #899, opened 2022-08-11
Show the image hash mismatch status for large files

paninaro:
Fixes https://github.com/Chia-Network/chia-blockchain-gui/issues/828

PR #902, opened 2022-08-12
Show the image hash mismatch status for large files (#899)

paninaro:
* Show the image hash mismatch status for large files

* More specific fix. Prior change wasn't matching on hash mismatches.

* Linter fix


seeden (approved):
lgtm

PR #905, opened 2022-08-12
Force legacy keyring migration. Legacy keyring support is now deprecated. (#904)

paninaro:
* Force legacy keyring migration. Legacy keyring support is now deprecated.

* Fixed empty passphrase / min length enforcement


seeden (approved):
lgtm

PR #944, opened 2022-08-17
added fee estimator component to gui

lipalong:
Added a component to the gui which displays fee estimations and which takes the fee input from the user


paninaro (approved):
approved so that we can continue building off of this.

seeden (dismissed):
lgtm

paninaro (approved):
lgtm

PR #945, opened 2022-08-17
Fix metadata decoding/parsing when content-type is missing

paninaro:
Electron's net.request has a long-standing bug wherein the IncomingMessage response object will not have a content-type header if the server originally responded with a redirect. The content-type header also appears to be missing in some cases for unknown reasons. This change works around the issue by reinterpreting the bytes as UTF-8 if the header is missing.


seeden (approved):
lgtm

PR #953, opened 2022-08-20
Fix method for identifying the NFT inbox (unassigned)

paninaro:
Without this fix, it's possible that the "Unassigned" NFT wallet for some users will have an associated DID.


seeden (approved):
lgtm

PR #951, opened 2022-08-19
fixed adding cats

seeden:
used own local storage hook because window listener is not working in electron for storage event
invalidate api-react cache
fixed wallet name and used meta name

PR #955, opened 2022-08-20
fixed adding cats (#951)

paninaro:
* used own local storage hook because window listener is not working in electron for storage event
invalidate api-react cache
fixed wallet name and used meta name

* added ability to rename and change visibility in one step


wallentx (approved):

seeden (approved):
lgtm

Issue #950, opened 2022-08-19
[Bug] Asset ID not shown when accepting offer


cameroncooper:
### What happened?

When accepting an offer for an Unknown CAT, I am told to verify the Asset ID but it is not shown.

![image](https://user-images.githubusercontent.com/898689/185700830-1179e8dd-0db5-493c-aefb-93b7600dd877.png)

### Version

1.5.1.dev452

### What platform are you using?

macOS

### What ui mode are you using?

GUI

paninaro:
Fix is in main. Needs to be included in 1.6.0

paninaro:
Fixed in 1.6.1 with new offer builder

PR #956, opened 2022-08-21
Fix tooltip in CAT offer rows #950

paninaro:
Fixes the issue where an unknown CAT's tooltip wouldn't show the asset ID.


seeden (approved):
lgtm

PR #958, opened 2022-08-22
fixed null values

seeden:
fixed non exists items in local storage


packages/core/src/hooks/useLocalStorage.ts
```diff
@@ -8,15 +8,15 @@ function getValueFromLocalStorage<T>(key: string, defaultValue: T) {
     // Get from local storage by key
```
ChiaMineJP, line 8: I think outer `try` block here is no longer necessary.

ChiaMineJP (changes requested):

packages/core/src/hooks/useLocalStorage.ts
seeden: thanks, removed

paninaro (approved):
lgtm

PR #974, opened 2022-08-27
Match Menu Theme with App Theme

zcomputerwiz:
Match the Menu theme to the App theme rather than the OS default.

### DIFF
--- a/packages/core/src/components/DarkModeToggle/DarkModeToggle.tsx
+++ b/packages/core/src/components/DarkModeToggle/DarkModeToggle.tsx
@@ -2,12 +2,17 @@ import React from 'react';
 import { IconButton } from '@mui/material';
 import { Brightness4, Brightness7 } from '@mui/icons-material';
 import useDarkMode from '../../hooks/useDarkMode';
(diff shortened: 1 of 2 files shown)


paninaro:
Thanks for the fix. @zcomputerwiz. I did a quick test and things look good.

One comment on the use of `window.require`... We've been moving away from `window.require` in favor of using `import { nativeTheme } from '@electron/remote';` to unify how modules are resolved. @seeden is there any harm in using an import in this case?

zcomputerwiz:
> Thanks for the fix. @zcomputerwiz. I did a quick test and things look good.
> 
> One comment on the use of `window.require`... We've been moving away from `window.require` in favor of using `import { nativeTheme } from '@electron/remote';` to standardize on how modules are resolved. @seeden is there any harm in using an import in this case?

I have pushed commits to use import { nativeTheme } from '@electron/remote'; instead as recommended, but haven't had a chance to test. I assume it will function as expected without additional changes?

paninaro (approved):
lgtm. Thanks for the contribution!

PR #975, opened 2022-08-29
NFT audio, video, model, document and unknown thumbnails

rifeljm:


packages/gui/src/components/nfts/detail/NFTDetailV2.tsx
```diff
@@ -21,34 +30,115 @@ import useNFTMetadata from '../../../hooks/useNFTMetadata';
...
     nftWallets.map((wallet: Wallet) => wallet.id),
   );
 
+  const [progressBarWidth, setProgressBarWidth] = React.useState(-1);
+  const [validated, setValidated] = React.useState(0);
+  const nftRef = React.useRef(null);
+
+  useEffect(() => {
+    validateSha256Remote(false); // false parameter means only validate files smaller than MAX_FILE_SIZE
+    ipcRenderer.on('sha256DownloadProgress', progressListener);
```
seeden, line 51: Please create a new hook which will use it instead of this complex event listening.
Hook can return { isLoading, progress, hash }  and use useEffect internally

packages/gui/src/components/nfts/detail/NFTDetailV2.tsx
```diff
@@ -21,34 +30,115 @@ import useNFTMetadata from '../../../hooks/useNFTMetadata';
...
   }, [nfts]);
 
-  const { metadata, isLoading: isLoadingMetadata } = useNFTMetadata(nft);
+  nftRef.current = nft;
+
+  function progressListener(_event, progressObject: any) {
+    const nft = nftRef.current;
+    if (
+      nft &&
+      nft.dataUris &&
```
seeden, line 72: You can use nft?.dataUris instead of nft && nft.dataUris

packages/gui/src/components/nfts/NFTPreview.tsx
```diff
@@ -162,22 +420,268 @@ export default function NFTPreview(props: NFTPreviewProps) {
         </body>
       </html>,
     );
-  }, [file, statusText, isStatusError]);
+  }, [file, statusText, isStatusError, thumbnail, error]);
+
+  function mimeType() {
+    const pathName: string = new URL(file).pathname;
+    return mime.lookup(pathName);
```
paninaro, line 427: Several of my NFTs don't have a `pathName` with a file extension. For those NFTs, this function returns `false` which then causes exceptions when trying to call `<false>.match(/whatever/)`.

As a workaround, I modified this function to return `undefined` if lookup returned `false`, and then updated all occurrences of `mimeType().` to `mimeType()?.`

That got me past the exceptions so that the gallery could render.

paninaro: Original exception:

```
Uncaught TypeError: mimeType(...).match is not a function
    at NFTPreview.tsx:394:27
    at updateMemo (react-dom.development.js:15867:1)
    at Object.useMemo (react-dom.development.js:16413:1)
    at useMemo (react.development.js:1532:1)
    at NFTPreview (NFTPreview.tsx:310:25)
    at renderWithHooks (react-dom.development.js:14985:1)
    at updateFunctionComponent (react-dom.development.js:17356:1)
    at beginWork (react-dom.development.js:19063:1)
    at HTMLUnknownElement.callCallback (react-dom.development.js:3945:1)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:3994:1)```

File: packages/gui/src/components/nfts/NFTCard.tsx [hunk 4]
```diff
@@ -7,7 +7,6 @@ import { MoreVert } from '@mui/icons-material';
 import styled from 'styled-components';
 import NFTPreview from './NFTPreview';
 import { type NFTInfo } from '@chia/api';
-import useNFTMetadata from '../../hooks/useNFTMetadata';
```
paninaro, line 10: This import is needed

packages/gui/src/components/nfts/NFTCard.tsx
```diff
@@ -41,7 +40,9 @@ export default function NFTCard(props: NFTCardProps) {
 
   const navigate = useNavigate();
 
-  const { metadata, isLoading } = useNFTMetadata(nft);
```
paninaro, line 44: This hook is needed. Without this, each card shows "Title Not Available"

packages/gui/src/components/nfts/NFTCard.tsx
```diff
@@ -41,7 +40,9 @@ export default function NFTCard(props: NFTCardProps) {
...
+  // const { metadata, isLoading } = useNFTMetadata(nft);
+  const isLoading: boolean = false;
+  const metadata: any = {};
```
paninaro, line 45: Can be removed once the `useNFTMetadata` hook is reinstated.

packages/gui/src/components/nfts/NFTDetails.tsx
```diff
@@ -231,6 +231,46 @@ export default function NFTDetails(props: NFTDetailsProps) {
       });
     }
 
+    if (metadata?.preview_image_uris) {
```
paninaro, line 234: Shouldn't the preview_image_uris/preview_video_uris support be in either NFTCard or NFTPreview?

packages/gui/src/components/nfts/NFTCard.tsx
seeden: It was just moved few lines down

packages/gui/src/components/nfts/NFTProgressBar.tsx
```diff
@@ -0,0 +1,25 @@
...
+    height: 10px;
+    border-radius: 2px;
+  }
+`;
+
+export default function NFTProgressBar({ percentage }) {
+  if (percentage === -1) return null;
+  return (
+    <ProgressBar>
+      <div style={{ width: `${percentage * 100}%` }}></div>
```
seeden, line 22: use <Box with sx property instead of direct style property and div component

packages/gui/src/electron/AudioThumbnailClass.ts
```diff
@@ -0,0 +1,87 @@
...
+/* ffmpeg library should not be packed into app.asar in production! */
+let pathToFfmpeg = ffmpegStatic.replace('app.asar', 'app.asar.unpacked');
+let pathToFfprobe = ffprobeStatic.path.replace('app.asar', 'app.asar.unpacked');
+
+const DEV = process.env.NODE_ENV !== 'production';
+/* ffmpeg binary remains in node_modules (development) */
+
+pathToFfmpeg = DEV
+  ? pathToFfmpeg.replace(
+      'build' + path.sep + 'electron',
```
seeden, line 16: use path.join instead of + operator

packages/gui/src/electron/main.tsx
```diff
@@ -29,14 +30,21 @@ import About from '../components/about/About';
 import packageJson from '../../package.json';
 import AppIcon from '../assets/img/chia64x64.png';
 import windowStateKeeper from 'electron-window-state';
+import validateSha256 from './validateSha256';
 
 const NET = 'mainnet';
 
 app.disableHardwareAcceleration();
+app.commandLine.appendSwitch('disable-http-cache');
```
seeden, line 38: why we are doing it?

packages/gui/src/electron/main.tsx
```diff
@@ -321,6 +329,7 @@ if (!handleSquirrelEvent()) {
           nodeIntegration: true,
           contextIsolation: false,
           nativeWindowOpen: true,
+          webSecurity: false,
```
seeden, line 332: Why you are disable webSecurity. If you have problem with CORS we can do it without disable of webSecurity

packages/gui/src/electron/validateSha256.ts
```diff
@@ -0,0 +1,68 @@
+import { net, ipcMain, IncomingMessage, BrowserWindow } from 'electron';
```
seeden, line 1: we already have implemented hash validation via hook useVerifyURIHash 
we should use same functionality 
if we need it in electron renderer then we can move everything from hook to ipcRenderer event

packages/gui/src/hooks/useNFTHash.ts
```diff
@@ -1,9 +1,82 @@
...
+  const metadataJson: any = useLoadingPreview(nft);
+
+  let { isValid, isLoading, thumbnail, error } = useVerifyHash(
+    uri,
+    metadataJson,
+    isPreview,
+    nft.dataHash,
+  );
+
+  thumbnail.type = isVideo(uri) ? 'video' : isAudio(uri) ? 'audio' : 'unknown';
```
seeden, line 66: we should add new enum for NFT type

packages/gui/src/hooks/useVerifyHash.ts
```diff
@@ -0,0 +1,147 @@
...
+} {
+  const [isValid, setIsValid] = useState(false);
+  const [isLoading, setIsLoading] = useState(false);
+  const [error, setError] = useState<string | undefined>();
+  const [thumbnail, setThumbnail] = useState({});
+
+  let lastError: any;
+
+  async function validateHash(metadataJson: any): Promise<void> {
+    let uris: string[] = [];
```
seeden, line 36: this method is too complicated.
use one try { setLoading(true); ... everything else } catch { setError(error) }  finally { setLoading (false)}

instead of using setError many times just throw new Error()

packages/gui/src/components/nfts/NFTDetails.tsx
paninaro: Disregard

packages/gui/src/electron/main.tsx
paninaro: We can disable the HTTP cache when our own caching is in place.

packages/gui/src/electron/main.tsx
paninaro: We can re-enable this. For reading from the FS, we can have the renderer process communicate with the main process via ipcRenderer. See `getRemoteFileContent`

PR #991, opened 2022-09-01
added support for auto-login

seeden:
added support for auto-login
fixed fingerprint value after login (api-react)


paninaro (approved):
lgtm. Tested and is working well.

Issue #733, opened 2022-05-13
[Bug] GUI shows error, but tx was submitted


mariano54:
### What happened?

When submitting a transaction, the wallet submits to multiple nodes.
The GUI then calls getTransaction, which will return the TX with all the results
Let's say one node responded but it was an error (NO_TRANSACTIONS_WHILE_SYNCING).
At this point, the GUI just displays the error. It appears the transaction failed. This caused me to retry, and i actually ended up making 2 identical transactions
Suggested fix: If getTransaction returns a transaction that has some errors, show the errors, but also show: Transaction will be retried, and might succeed in the future
also we can retry for several seconds in case another node responds and sets the status

### Version

main

### What platform are you using?

Linux

### What ui mode are you using?

GUI

mariano54:
Any progress on this @seeden ?

mariano54:
Bump @seeden

seeden:
Fixed via https://github.com/Chia-Network/chia-blockchain-gui/pull/1004

PR #1004, opened 2022-09-07
[Bug] GUI shows error, but tx was submitted #733

seeden:
Fixed [Bug] GUI shows error, but tx was submitted #733
Waiting for correct transaction and ignoring transaction with error message NO_TRANSACTIONS_WHILE_SYNCING


paninaro (approved):
lgtm

PR #1014, opened 2022-09-12
Enable webpack fs caching to speed up launching via 'npm run dev:fast'

paninaro:
Relaunching takes ~12 seconds now compared to ~1:30 previously. I haven't noticed any increased build times as a result of this change.


seeden (approved):
lgtm

Issue #965, opened 2022-08-25
Update to Electron 20 - Take 2


paninaro:
- [x] Update to latest Electron 20
- [x] Remove com.apple.security.cs.allow-unsigned-executable-memory from macOS entitlements
- [x] Add com.apple.security.cs.allow-jit

References:
https://www.electronjs.org/docs/latest/tutorial/code-signing
https://github.com/electron-userland/electron-builder/blob/master/packages/app-builder-lib/templates/entitlements.mac.plist
https://github.com/electron/electron-notarize#prerequisites

paninaro:
PR: https://github.com/Chia-Network/chia-blockchain-gui/pull/1015

PR #1015, opened 2022-09-13
Bump Electron to 20.1.3 and update entitlements based on latest guidance

paninaro:
Added "com.apple.security.cs.allow-jit" entitlement to fix crash on launch on M1 Macs.
Removed "com.apple.security.cs.allow-unsigned-executable-memory" entitlement based on the following guidance:

"Note that up until Electron 12, the com.apple.security.cs.allow-unsigned-executable-memory entitlement was required as well. However, it should not be used anymore if it can be avoided." (https://www.electronjs.org/docs/latest/tutorial/code-signing)

Tested on Intel and M1 using installers:
https://download.chia.net/dev/Chia-1.5.2-dev2571-41f97d3b.dmg
https://download.chia.net/dev/Chia-1.5.2-dev2571-41f97d3b-arm64.dmg


paninaro:
Addresses GH issue https://github.com/Chia-Network/chia-blockchain-gui/issues/965

ChiaMineJP (approved):
lgtm.
I confirmed that installing/running https://download.chia.net/dev/Chia-1.5.2-dev2571-41f97d3b-arm64.dmg on my m1 Mac worked without problems.

PR #1017, opened 2022-09-13
renaming keys

seeden:
first working version renaming keys


packages/api-react/src/services/daemon.ts
```diff
@@ -1,23 +1,100 @@
...
+      }
+    >({
+      query: ({ fingerprint, label }) => ({
+        command: 'setLabel',
+        service: Daemon,
+        args: [fingerprint, label],
+      }),
+      invalidatesTags: () => ['DaemonKey'],
+      transformResponse: (response: any) => {
+        console.log('setLabel response', response);
```
paninaro, line 66: Remove?

packages/api-react/src/services/daemon.ts
```diff
@@ -1,23 +1,100 @@
...
+      }
+    >({
+      query: ({ fingerprint }) => ({
+        command: 'deleteLabel',
+        service: Daemon,
+        args: [fingerprint, label],
+      }),
+      invalidatesTags: () => ['DaemonKey'],
+      transformResponse: (response: any) => {
+        console.log('deleteLabel response', response);
```
paninaro, line 84: Remove?

packages/api-react/src/services/daemon.ts
```diff
@@ -1,23 +1,100 @@
...
+    deleteLabel: build.mutation<
+      boolean,
+      {
+        fingerprint: number;
+      }
+    >({
+      query: ({ fingerprint }) => ({
+        command: 'deleteLabel',
+        service: Daemon,
+        args: [fingerprint, label],
```
paninaro, line 80: Remove `label` argument?

packages/api-react/src/services/wallet.ts
```diff
@@ -600,6 +601,9 @@ export const walletApi = apiWithTag.injectEndpoints({
       }),
       invalidatesTags: (_result, _error, { fingerprint }) => [
         { type: 'Keys', id: fingerprint },
+        { type: 'Keys', id: 'LIST' },
+        { type: 'DaemonKey', id: fingerprint },
+        { type: 'DaemonKey', id: 'LIST' },
```
paninaro, line 606: If these are necessary, should the same be applied to the `addKey` and `deleteAllKeys` mutations?

packages/core/package.json
```diff
@@ -83,25 +88,30 @@
     "babel-plugin-styled-components": "2.0.2",
     "babel-plugin-transform-imports": "2.0.0",
     "cross-env": "7.0.3",
+    "cypress": "10.8.0",
+    "electron": "19.0.4",
```
paninaro, line 92: I believe this is what Snyk was complaining about. If it's necessary, we can probably use 20.1.3 like we do in `packages/gui/package.json`

packages/core/package.json
```diff
@@ -83,25 +88,30 @@
...
     "@chia/api-react": "1.0.0",
+    "@electron/remote": "2.0.8",
     "@lingui/core": "3.13.0",
     "@lingui/react": "3.13.0",
     "@mui/icons-material": "5.5.1",
     "@mui/lab": "5.0.0-alpha.75",
     "@mui/material": "5.5.2",
     "@mui/styled-engine-sc": "5.5.2",
     "@mui/styles": "5.5.1",
+    "electron": ">=19.0.0",
```
paninaro, line 114: Maybe this too

packages/core/src/screens/SelectKey/SelectKeyRenameForm.tsx
```diff
@@ -0,0 +1,109 @@
...
+        <TextField
+          name="label"
+          size="small"
+          disabled={!canSubmit}
+          data-testid="SelectKeyRenameForm-label"
+          onKeyDown={handleKeyDown}
+          InputProps={{
+            endAdornment: (
+              <InputAdornment position="end">
+                <Tooltip title={<Trans>Cancel Editation</Trans>}>
```
paninaro, line 79: Maybe just "Cancel" by itself?

packages/gui/playwright-report/index.html
```diff
@@ -0,0 +1,62 @@
+
```
paninaro, line 1: This file was deleted in main. Do we need it? I was thinking we should just add it to .gitignore if it keeps coming back.

paninaro:
Were the cypress inclusions intentional? If it's simple enough, I would probably put those in a separate branch/PR, but if not we can just keep it included in this PR.

packages/core/package.json
seeden: Rebased

packages/core/package.json
seeden: Updated to 20.1.3

packages/api-react/src/services/wallet.ts
seeden: Nice catch, updated.

seeden:
> Were the cypress inclusions intentional? If it's simple enough, I would probably put those in a separate branch/PR, but if not we can just keep it included in this PR.

I will create separate PR. Cypress was removed

paninaro (approved):
lgtm

PR #1020, opened 2022-09-14
Fix incorrect display of assets when viewing offers created by self

paninaro:
When a user views an NFT offer that they created from the "Offers you created" list, the offered/requested assets are reversed. This fixes the issue by flipping the offered/requested assets in the above scenario.

Doesn't impact "imported" NFT offers (viewing from a blob/file).
Doesn't impact token offers as those use a separate component.


paninaro:
image

### Without fix:

image

### With fix:

image

### When viewing offer blob/file (imported == true)

image

seeden (approved):
lgtm

PR #1022, opened 2022-09-15
Add jest and 'test:jest' script

paninaro:
Deleted old/broken test files


packages/gui/package.json
```diff
@@ -106,6 +106,12 @@
       ]
     }
   },
+  "jest": {
+    "testPathIgnorePatterns": [
+      "/node_modules/",
+      "<rootDir>/tests/"
```
paninaro, line 112: Added because this is where the Playwright tests currently reside. We may want to move those somewhere else.

ChiaMineJP:
Do you want to use both jest and playwright?

paninaro:
> Do you want to use both jest and playwright?

Yes, @jfawChia is working on writing end-to-end tests using Playwright, while I'd like to use jest for unit tests. I'm open to other suggestions as well.

seeden (approved):
lgtm

Issue #1019, opened 2022-09-14
[BUG] t.filter is not a function


karmiller:
### What happened?

Hi I have 3 NFTs within the same collect that that every time i click the GUI errors with the below and the wallet wants me to reload the app.    Something went wrong
Error: t.filter is not a function
at webpack://@chia/gui/src/components/nfts/NFTRankings.tsx:70:23 NFTRankings
at webpack://@chia/gui/node_modules/react-dom/cjs/react-dom.production.min.js:172:114 useLayoutEffect
at webpack://@chia/gui/node_modules/react/cjs/react.production.min.js:23:112 useLayoutEffect
at webpack://@chia/gui/src/components/nfts/NFTRankings.tsx:69:29 NFTRankings
at webpack://@chia/gui/node_modules/react-dom/cjs/react-dom.production.min.js:157:136 Ch
at webpack://@chia/gui/node_modules/react-dom/cjs/react-dom.production.min.js:267:459 b?b.sibling:null}I
at webpack://@chia/gui/node_modules/react-dom/cjs/react-dom.production.min.js:250:346 Pj
at webpack://@chia/gui/node_modules/react-dom/cjs/react-dom.production.min.js:250:277 Pj
at webpack://@chia/gui/node_modules/react-dom/cjs/react-dom.production.min.js:250:137 Pj
at webpack://@chia/gui/node_modules/react-dom/cjs/react-dom.production.min.js:243:162 Lj

### Version

15.1 but tried in 14 and 15

### What platform are you using?

Windows

### What ui mode are you using?

GUI

### Relevant log output or stacktrace

```shell
t.filter is not a function

URL
#/dashboard/nfts/nft1h574vc6ujs3luu5yn3670httdwep2nqkz4zuk2qadv8ge5ppgqrqguejmc

Stacktrace
at webpack://@chia/gui/src/components/nfts/NFTRankings.tsx:70:23 NFTRankings
at webpack://@chia/gui/node_modules/react-dom/cjs/react-dom.production.min.js:172:114 useLayoutEffect
at webpack://@chia/gui/node_modules/react/cjs/react.production.min.js:23:112 useLayoutEffect
at webpack://@chia/gui/src/components/nfts/NFTRankings.tsx:69:29 NFTRankings
at webpack://@chia/gui/node_modules/react-dom/cjs/react-dom.production.min.js:157:136 Ch
[...]

PR #1023, opened 2022-09-15
Fix crash when viewing an NFT with malformed attributes

paninaro:
Fixes issue #1019

In the bug report, the NFT has `attributes` to set `{trait_type: 'un1qu3EXE', value: 'm32k13 7233'}` instead of being an array of objects.

### DIFF
--- a/packages/gui/src/components/nfts/NFTProperties.tsx
+++ b/packages/gui/src/components/nfts/NFTProperties.tsx
@@ -87,7 +87,10 @@ export default function NFTProperties(props: NFTPropertiesProps) {
   const { attributes } = props;
 
   const valueAttributes = useMemo(() => {
-    return attributes?.filter((attribute) => !isRankingAttribute(attribute));
+    if (Array.isArray(attributes)) {
+      return attributes.filter((attribute) => !isRankingAttribute(attribute));
+    }
+    return [];
   }, [attributes]);
 
   if (!valueAttributes?.length) {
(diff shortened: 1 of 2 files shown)


packages/gui/src/components/nfts/NFTProperties.tsx
```diff
@@ -87,7 +87,10 @@ export default function NFTProperties(props: NFTPropertiesProps) {
   const { attributes } = props;
 
   const valueAttributes = useMemo(() => {
-    return attributes?.filter((attribute) => !isRankingAttribute(attribute));
+    if (Array.isArray(attributes)) {
```
seeden, line 90: you can use when attributes === undefined

return attributes?.filter((attribute) => !isRankingAttribute(attribute)) ?? [];

or when attributes is null

return attributes?.filter((attribute) => !isRankingAttribute(attribute)) || [];

instead

packages/gui/src/components/nfts/NFTRankings.tsx
```diff
@@ -67,7 +67,10 @@ export default function NFTRankings(props: NFTRankingsProps) {
   const { attributes } = props;
 
   const rankingsAttributes = useMemo(() => {
-    return attributes?.filter(isRankingAttribute);
+    if (Array.isArray(attributes)) {
```
seeden, line 70: same here

packages/gui/src/components/nfts/NFTProperties.tsx
paninaro: So the issue wasn't that attributes was undefined, but rather, it was an object instead of an array. i.e.
`{trait_type: 'un1qu3EXE', value: 'm32k13 7233'}`

instead of:

`[{trait_type: 'un1qu3EXE', value: 'm32k13 7233'}]`

The failure occurred because `filter()` was being called on an object

seeden: I see.

seeden (approved):
lgtm

PR #1031, opened 2022-09-26
Playwright 1 6 2nd

jfawChia:
All scripts have been updated to reflect 1.6 features, Passphrase login, and wallet sync function.


paninaro:
@jfawChia We should revert the modified messages.po and messages.js files in your branch. These files are modified when you run `npm run build`, but we shouldn't be checking them in normally (we need a better process for this).

Can you run this command line to restore those files to their previous version in main:
```git --no-pager diff --name-only main | grep -e messages\.po -e messages\.js | xargs git checkout main```

You should see the following output or something similar:
```Updated 193 paths from 6ba9d040f```

You can then run `git add packages` to stage those reverted files, and then `git commit -m 'Reverted messages.[po,js]'`.

File: .gitignore [hunk 1]
```diff
@@ -100,3 +100,9 @@ packages/**/playwright-report/index.html
 
 .env
 dist
+<<<<<<< HEAD
+
+# playwright non-essentials
+packages/gui/playwright-report
+=======
+>>>>>>> main
```
paninaro, line 108: Looks like there are some git merge conflict markers left in this file.

Since we already have `packages/**/playwright-report/index.html` on line 99, I think we can just delete lines 103-108

paninaro (approved):

PR #1034, opened 2022-09-27
replaced is_running with running_services

seeden:
used running_services inside useService hook and stopped to use all useIsRunning hooks

### DIFF
--- a/packages/api-react/src/hooks/useService.ts
+++ b/packages/api-react/src/hooks/useService.ts
@@ -2,8 +2,8 @@ import { useEffect, useState } from 'react';
 import { ServiceName } from '@chia/api';
 import { useClientStartServiceMutation } from '../services/client';
 import {
-  useIsServiceRunningQuery,
   useStopServiceMutation,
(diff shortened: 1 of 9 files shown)


packages/api-react/src/hooks/useService.ts
```diff
@@ -35,14 +36,12 @@ export default function useService(
...
     refetch,
     error,
-  } = useIsServiceRunningQuery(
-    {
-      service,
-    },
+  } = useRunningServicesQuery(
+    {},
     {
       pollingInterval: 1000,
```
paninaro, line 46: What do you think about increasing this value? Maybe at startup/shutdown we could have it poll more frequently (like it is now), but otherwise when operating normally, only poll every 5-10 seconds or so?

seeden: nice improvement I will add it :)

seeden: added

paninaro (approved):
lgtm

PR #1035, opened 2022-09-27
Spacescan.io offer sharing for token/NFT offers

paninaro:


packages/gui/src/components/offers/OfferShareDialog.tsx
```diff
@@ -862,6 +917,87 @@ OfferShareHashgreenDialog.defaultProps = {
...
+  const openExternal = useOpenExternal();
+  const [sharedURL, setSharedURL] = React.useState('');
+
+  function handleClose() {
+    onClose(false);
+  }
+
+  async function handleConfirm() {
+    const url = await postToSpacescan(offerData, testnet);
+    console.log(`Spacescan.io URL: ${url}`);
```
seeden, line 931: for logging we are using "debug" npm library. you can use it or remove console.log

packages/gui/src/components/offers/OfferShareDialog.tsx
```diff
@@ -862,6 +917,87 @@ OfferShareHashgreenDialog.defaultProps = {
...
+      title={<Trans>Share on Spacescan.io</Trans>}
+      onConfirm={handleConfirm}
+      open={open}
+      onClose={onClose}
+    />
+  );
+}
+
+OfferShareSpacescanDialog.defaultProps = {
+  open: false,
```
seeden, line 997: In the future please use typescript default properties directly inside the component. This syntax is deprecated because it has several disadvantages and typescript does not know default values after that

packages/gui/src/components/offers/OfferShareDialog.tsx
paninaro: Thanks, I added a new `chia-gui:offers` log outlet. I'd like to start doing more GUI logging, so this is a good reminder.

packages/gui/src/components/offers/OfferShareDialog.tsx
paninaro: Updated

seeden (approved):
lgtm

Issue #1010, opened 2022-09-09
GUI: When naming a stray CAT wallet, don't add wallet until switch is turned on


paninaro:
Users who have many CAT wallets may want to give those wallets proper names without actually adding the wallet.

Currently, the UI will create the wallet with the provided name, even if the user hasn't turned the wallet's visibility toggle on.

Proposed solution is to have the UI track the names of those CAT wallets that haven't been added yet. This would allow us to display the user-provided name without actually adding the wallet in the backend (which incurs sync/resource overhead). If the user later decides to show the wallet, the wallet would be created as needed.

seeden:
Created PR https://github.com/Chia-Network/chia-blockchain-gui/pull/1044

seeden:
PR merged. Closing

PR #1044, opened 2022-10-03
added support for local (localStorage) CAT names

seeden:
[GUI: When naming a stray CAT wallet, don't add wallet until switch is turned on#1010](https://github.com/Chia-Network/chia-blockchain-gui/issues/1010)


packages/api-react/src/hooks/useLocalStorage.ts
```diff
@@ -3,26 +3,32 @@ import EventEmitter from '../utils/EventEmitter';
...
-function getValueFromLocalStorage<T>(key: string, defaultValue: T) {
+function getValueFromLocalStorage<T>(
+  key: string,
+  defaultValue?: T
+): T | undefined {
   const item = window.localStorage.getItem(key);
 
   if (item === undefined || item === null) {
-    return defaultValue;
+    return undefined;
```
paninaro, line 13: This change appears to be causing an exception when loading WalletsSidebar with the useHiddenWallet hook.

I'm seeing this when selecting a key at launch (auto-login disabled):

image

paninaro: The issue seems to be that hiddenFingerprints is undefined:

image

paninaro: Seems to work if I add:

`
if (!hiddenWalletids) {
  return false;
}
`

paninaro: Hiding a wallet seems to throw an exception as well

seeden: I fixed typo in my code. Thanks for nice catch

paninaro (approved):
lgtm

PR #1053, opened 2022-10-06
Ll.fee estimator aug8

lipalong:
Added estimated fee input component with dropdown selection and custom input options.


lipalong:
When the fee estimates refresh, if the user has already selected a fee/time combo, the new updated fee which corresponds to their selected time is shown correctly in the fee input box. However, the old fee value that they selected is the one which gets returned instead of the updated fee value (e.g. the fee value that gets used in WalletSend.tsx when sending a transaction). This needs to be fixed – any ideas?

Issue #1061, opened 2022-10-17
[Bug] When I rename two wallets with the same name, I get a cryptic message. 


sethjenks:
### What happened?

When I rename two wallets with the same name, I get a cryptic message. "Error malformed request"

I am also getting the same error when I am in the edit mode and don't make any changes to the text and try to save. 

<img width="719" alt="CleanShot 2022-10-17 at 14 53 33@2x" src="https://user-images.githubusercontent.com/663017/196281095-911e89cb-682a-4907-b0ca-d075763c4c99.png">

It would be best if we just allowed people to make duplicate names.

It could also say, "Hold up! Wallets must have unique names". This would be annoying in the case above though where I have a unique name already and I can't save it. 

### Version

1.6.1b4-dev16

### What platform are you using?

macOS

### What ui mode are you using?

GUI

emlowe:
Fixed in 1.6.1

PR #1064, opened 2022-10-18
Show error details when an RPC fails, if available

paninaro:
Don't call setLabel if the new label matches the old label

Addresses #1061

### DIFF
--- a/packages/api/src/Client.ts
+++ b/packages/api/src/Client.ts
@@ -27,10 +27,13 @@ export default class Client extends EventEmitter {
   private ws: any;
 
   private connected = false;
-  private requests: Map<string, {
-    resolve: (value: Response) => void;
-    reject: (reason: Error) => void;
-  }> = new Map();
+  private requests: Map<
+    string,
+    {
+      resolve: (value: Response) => void;
+      reject: (reason: Error) => void;
+    }
+  > = new Map();
 
   private services: Set<ServiceName> = new Set();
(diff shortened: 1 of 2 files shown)


packages/api/src/Client.ts
```diff
@@ -323,20 +338,31 @@ export default class Client extends EventEmitter {
...
+            '[Error 13] Permission denied. You are trying to access a file/directory without having the necessary permissions. Most likely one of the plot folders in your config.yaml has an issue.';
         } else if (errorMessage == '22') {
-          errorMessage = '[Error 22] File not found. Most likely one of the plot folders in your config.yaml has an issue.';
+          errorMessage =
+            '[Error 22] File not found. Most likely one of the plot folders in your config.yaml has an issue.';
         }
 
         log(`Request ${requestId} rejected`, errorMessage);
 
+        if (message?.data?.errorDetails?.message) {
```
seeden, line 350: Maybe we should use "else if" instead just "if".
Second option is to use this message directly on line 338.

seeden (changes requested):

packages/api/src/Client.ts
paninaro: Moved it into an else if grouped with the logic right before. The error message is set in this case, but there's also error details. So the idea was to combine both into a single message to the user. This form of error message is only used by the keyring backend code currently.

seeden (approved):
lgtm

PR #1069, opened 2022-10-18
fix displayed currency code in fee estimator

lipalong:


paninaro (approved):
Thanks for the quick fix! Tested and confirmed working.

Issue #1068, opened 2022-10-09
Wrong graph for CH21, Spacebucks and some other CATs


korolevsv:
### What happened?

Graph for CH21, Spacebucks and some other CATs look not like a line but as a unit impulse (delta-function)

### Version

1.6.1b3
<img width="1232" alt="Wrong graph for CH21" src="https://user-images.githubusercontent.com/8134234/195174386-d4a3cf80-2e95-45e1-ac0c-de6cf1accc48.png">

### What platform are you using?

macOS

### What ui mode are you using?

GUI

korolevsv:
The bug still remained in the 1.6.1-rc1 :-(

seeden:
I also fixed spike graph in https://github.com/Chia-Network/chia-blockchain-gui/pull/1089

Issue #1065, opened 2022-10-18
[BUG] Cannot read properties of undefined (reading 'toUpperCase')


korolevsv:
### What happened?

GUI stuck while plotting Bladebit. Temp directory -342 GB on disk, 3080 files. Free space on disk only 2 GB left. No changes after 10 hours.
Error occured when switched several times between different GUI tabs (Pols-Wallet-Full Node)

### Version

1.6.1b3

### What platform are you using?

Windows

### What ui mode are you using?

GUI

### Relevant log output or stacktrace

```shell
Cannot read properties of undefined (reading 'toUpperCase')

URL
#/dashboard/wallets/7

Stacktrace
at webpack://@chia/wallets/dist/esm.js:800:67 labels
at webpack://@chia/gui/node_modules/victory-core/es/victory-util/helpers.js:169:29 evaluateProp
at webpack://@chia/gui/node_modules/victory-tooltip/es/victory-tooltip.js:136:17 value
at webpack://@chia/gui/node_modules/victory-tooltip/es/victory-tooltip.js:492:32 Helpers.evaluateProp
at webpack://@chia/gui/node_modules/victory-tooltip/es/victory-tooltip.js:508:18 value
at webpack://@chia/gui/node_modules/react-dom/cjs/react-dom.production.min.js:187:187 qi
at webpack://@chia/gui/node_modules/react-dom/cjs/react-dom.production.min.js:186:172 pi
at webpack://@chia/gui/node_modules/react-dom/cjs/react-dom.production.min.js:269:426 a=b.pendingProps;f=e._init;e=f(e._payload);b.type=e;f=b.tag=hk(e);a=lg(e,a);switch
at webpack://@chia/gui/node_modules/react-dom/cjs/react-dom.production.min.js:250:346 Pj
at webpack://@chia/gui/node_modules/react-dom/cjs/react-dom.production.min.js:250:277 Pj
```

paninaro:
Fix has been included in 1.6.1

PR #1072, opened 2022-10-20
Fixed wallet balance graph bugs #1068

paninaro:
* The graph units were assumed to be XCH. The wallet type and name are now passed in to handle CATs.
* Fixed the graph appearance when only a single transaction was present.

Should also address: #1065


seeden (approved):
lgtm

PR #1073, opened 2022-10-20
fixed fee estimator progress bar early reset issue

lipalong:
Fixed an issue where the fee estimator's progress bar was resetting to zero too early.


paninaro (approved):
lgtm

PR #1075, opened 2022-10-20
Show error details when an RPC fails, if available (#1064)

paninaro:
* Show error details when an RPC fails, if available Don't call setLabel if the new label matches the old label

* Moved detailed error msg handling into else if block


seeden (approved):
lgtm

PR #1079, opened 2022-10-21
Nft gallery hide objectionable content

rifeljm:


packages/gui/src/components/nfts/NFTPreview.tsx
```diff
@@ -538,17 +557,46 @@ export default function NFTPreview(props: NFTPreviewProps) {
...
+          <script></script>
         </head>
         <body>{mediaElement}</body>
       </html>,
     );
 
+    const script = `
+        window.addEventListener("DOMContentLoaded", () => {
+          document.querySelector('video').addEventListener('mouseenter', function() {
+            console.log('window.parent', window.parent);
```
paninaro, line 591: Remove

packages/gui/src/components/nfts/NFTPreview.tsx
```diff
@@ -538,17 +557,46 @@ export default function NFTPreview(props: NFTPreviewProps) {
...
+          <source src={thumbnail.binary || file} />
+        </audio>
+      );
+    }
+
     let elem = renderToString(
       <html>
         <head>
           <style dangerouslySetInnerHTML={{ __html: style }} />
+          <script></script>
```
paninaro, line 582: Remove?

packages/gui/src/components/nfts/gallery/NFTGallery.tsx
```diff
@@ -65,6 +65,8 @@ export default function NFTGallery() {
     defaultCacheSizeLimit,
   );
 
+  const [sensitiveContentObject] = useLocalStorage('sensitive-content');
```
paninaro, line 68: Needs a default, otherwise sensitiveContentObject is undefined and leads to an exception on line 95.

```suggestion
  const [sensitiveContentObject] = useLocalStorage('sensitive-content', {});
```

packages/gui/src/components/nfts/gallery/NFTGallery.tsx
```diff
@@ -90,6 +92,10 @@ export default function NFTGallery() {
         return false;
       }
 
+      if (hideObjectionableContent && sensitiveContentObject[nft.$nftId]) {
```
paninaro, line 95: Does `sensitiveContentObject` need to be included in the useEffect dependency array?

In my testing, I'm seeing `sensitive-content: true` NFTs appearing in the gallery when I have 'Hide objectionable content' set to true. If I set it to false, and then back to true, the gallery hides the `sensitive-content: true` NFT.

I can repro this easily by having the gallery open and minting an NFT with sensitive-content set to true. The NFT will appear in the gallery when it should be hidden.

paninaro (changes requested):
The iframe changes look fine, but hiding sensitive content isn't working properly for me when an NFT appears in the gallery for the first time.

paninaro (dismissed):

paninaro (approved):

Issue #1077, opened 2022-10-20
GUI: Viewing offer in offer builder fails if offered/requested CAT wallet doesn't already exist


paninaro:
Viewing an offer for a CAT that you don't have a wallet for leads to the following:

![Image](https://user-images.githubusercontent.com/339312/197065195-537592f6-f436-44e1-b5ae-095e7a917c6f.png)

Instead, the offer should be displayed, and the unknown CATs should be labeled as such with a warning about accepting the offer.

paninaro:
Fixed in 1.6.1

PR #1080, opened 2022-10-22
Offer builder fixes

paninaro:
Allow accepting an offer for unknown CATs
Display warnings when unknown CATs are present in an offer Create Offer wallet entry points now launch Offer Builder instead of the old editor Don't allow fee editing when viewing a pending/accepted offer Fixed exception being raised when pasting bad data into the offer importer Fixed display of royalties in the offer builder when a fungible amount wasn't set Added link to taildatabase for token offers

Addresses https://github.com/Chia-Network/chia-blockchain-gui/issues/1077


seeden (approved):
lgtm

PR #1081, opened 2022-10-23
Fixed missing wallet id property on NFTs in the details screen.

paninaro:
This was causing some operations to fail, like burning or transferring.


seeden (approved):
lgtm

PR #1090, opened 2022-10-24
One-sided offers should not be possible [Bug] #1082

seeden:
Fix for One-sided offers should not be possible [Bug] #1082

Use the offer builder to make an offer where only one side has any assets. For example,

1 XCH in exchange for nothing
nothing in exchange for 1 XCH

This shouldn't be allowed


paninaro (approved):
lgtm

PR #1093, opened 2022-10-24
Merge nft-gallery-hide-objectionable-content

paninaro:
Previously merged into main. This merges the same changes into release/1.6.1.

Includes changes to move audio/video tags into the iframe when displaying NFT content, as well as support for auto-hiding NFTs with sensitive-content set to true.


seeden (approved):
lgtm

PR #1112, opened 2022-10-27
Fix useLocalStorage imports to point to api-react instead of core

paninaro:


danieljperry (commented):
This fixes the issue for me.

seeden (approved):
lgtm

PR #1122, opened 2022-10-29
Tighten up sensitive_content checking

paninaro:
Fixed issue where benign NFTs could be flagged as sensitive_content

Metadata with `"sensitive_content" : "false"` was being flagged as sensitive content. We now check for a boolean value of true, or a string of 'true'.

Additionally, the `setSensitiveContent()` nested function was incorrectly assuming that the passed in data was a string, instead of an already parsed JSON object. The captured values inside of the nested function were also not necessarily the expected values, so `setSensitiveContent()` now takes the `nftId` as a function param.


seeden (approved):
lgtm

PR #1123, opened 2022-10-31
Show royalty amounts and totals when viewing an offer

paninaro:
Show the breakdown of royalties for each CAT/XCH amount in the offer
Fixed fee section expansion when viewing an offer
Fixed tooltip copy-to-clipboard functionality in offer builder values
Show NFT minter name/DID

### DIFF
--- a/packages/gui/src/components/offers2/OfferBuilderNFT.tsx
+++ b/packages/gui/src/components/offers2/OfferBuilderNFT.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { ReactNode } from 'react';
 import { Trans } from '@lingui/macro';
 import { Flex, Loading } from '@chia/core';
 import { useGetNFTInfoQuery } from '@chia/api-react';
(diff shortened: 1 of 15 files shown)


packages/gui/src/components/offers2/OfferBuilderNFT.tsx
```diff
@@ -29,7 +30,7 @@ function PreviewCard(props) {
   );
 }
 
-export type OfferBuilderNFTProps = {
+export type OfferBuilderNFTProps = ReactNode & {
```
seeden, line 33: ReactNode should not be here

packages/gui/src/components/offers2/OfferBuilderToken.tsx
```diff
@@ -7,15 +7,24 @@ import { useWatch } from 'react-hook-form';
 import OfferBuilderValue from './OfferBuilderValue';
 import OfferBuilderWalletAmount from './OfferBuilderWalletAmount';
 
-export type OfferBuilderTokenProps = {
+export type OfferBuilderTokenProps = ReactNode & {
```
seeden, line 10: Please remove ReactNode from here

seeden (approved):
lgtm

PR #1140, opened 2022-11-01
Show royalties on both sides of the offer builder.

paninaro:
Added a shadow to NFT status pill text for better visibility.
Show "Unknown" instead of unknown CAT asset id in offer builder.


seeden (approved):
lgtm

PR #1149, opened 2022-11-03
Tests to cover toBech32m functionality

paninaro:
toBech32m will now throw if the input hex string is invalid (non-hex chars)
decodeBech32m now converts the decoded data to a utf8 string by default (not currently used)

100% test coverage of code reached by calculateRoyalties


seeden (approved):
lgtm

Issue #1143, opened 2022-11-02
Error message with trailing spaces in Wallet name


stewari5:
### What happened?

Got an error message when I created a new wallet and named it with trailing spaces in the name. The wallet was still created but name was the default name

![image](https://user-images.githubusercontent.com/107882139/199406238-b1836550-15e8-41e6-b1b3-cf7a3ef36b45.png)

### Version

1.6.1 RC5

### What platform are you using?

Windows

### What ui mode are you using?

GUI

paninaro:
We should use the new `add_private_key` RPC added to the daemon for this. I've prepared a PR to make a necessary addition to that RPC: https://github.com/Chia-Network/chia-blockchain/pull/13814

paninaro:
Fixed with https://github.com/Chia-Network/chia-blockchain-gui/pull/1157

paninaro:
### Testing Notes

This affected both the creation of a newly generated key as well as importing from a 24-word mnemonic.

### Steps to repro
1. Click "Create a New Private Key" or "Import from Mnemonic (24 Words)"
2. If importing from a mnemonic, enter the mnemonic
3. Enter a name for the new wallet key, and include spaces at the end of the name
4. Click "Next"

### Observe
Prior to the fix, an error dialog would appear indicating that the key already existed.

PR #1157, opened 2022-11-04
Use daemon add_private_key RPC instead of old add_key wallet RPC

paninaro:
Strip whitespace when adding/importing from mnemonic phrase

Fixes https://github.com/Chia-Network/chia-blockchain-gui/issues/1143

### DIFF
--- a/packages/wallets/src/components/WalletAdd.tsx
+++ b/packages/wallets/src/components/WalletAdd.tsx
@@ -9,9 +9,8 @@ import {
 import { useForm } from 'react-hook-form';
 import {
   useGenerateMnemonicMutation,
-  useAddKeyMutation,
+  useAddPrivateKeyMutation,
   useLogInMutation,
-  useSetLabelMutation,
 } from '@chia/api-react';
 import { useNavigate } from 'react-router';
 import { useEffectOnce } from 'react-use';
@@ -33,8 +32,7 @@ export default function WalletAdd() {
   const navigate = useNavigate();
   const [generateMnemonic, { data: words, isLoading }] =
     useGenerateMnemonicMutation();
-  const [setLabel] = useSetLabelMutation();
-  const [addKey] = useAddKeyMutation();
+  const [addKey] = useAddPrivateKeyMutation();
   const [logIn] = useLogInMutation();
   const methods = useForm<FormData>({
     defaultValues: {
@@ -61,16 +59,9 @@ export default function WalletAdd() {
 
     try {
       const fingerprint = await addKey({
-        mnemonic: words,
-        type: 'new_wallet',
+        mnemonic: words.join(' '),
+        ...(label && { label: label.trim() }),
       }).unwrap();
-
-      if (label) {
-        await setLabel({
-          fingerprint,
-          label,
-        }).unwrap();
-      }
 
       await logIn({
         fingerprint,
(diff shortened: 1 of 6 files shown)


packages/wallets/src/components/WalletImport.tsx
```diff
@@ -116,17 +111,10 @@ export default function WalletImport() {
     }
 
     const fingerprint = await addKey({
-      mnemonic: mnemonicWords,
-      type: 'new_wallet',
+      mnemonic: mnemonicWords.join(' '),
+      ...(label && { label: label.trim() }),
```
ChiaMineJP, line 115: The same as above

packages/wallets/src/components/WalletAdd.tsx
```diff
@@ -61,17 +59,10 @@ export default function WalletAdd() {
 
     try {
       const fingerprint = await addKey({
-        mnemonic: words,
-        type: 'new_wallet',
+        mnemonic: words.join(' '),
+        ...(label && { label: label.trim() }),
```
ChiaMineJP, line 63: Surprisingly `{...""}` and `{...false}` works but I rather prefer
```javascript
label: label.trim() ? label.trim() : undefined,
```
because it is more clear to me.

packages/wallets/src/components/WalletImport.tsx
```diff
@@ -44,8 +40,7 @@ type FormData = {
 
 export default function WalletImport() {
   const navigate = useNavigate();
-  const [setLabel] = useSetLabelMutation();
-  const [addKey] = useAddKeyMutation();
+  const [addKey] = useAddPrivateKeyMutation();
```
ChiaMineJP, line 43: The same as above.

packages/wallets/src/components/WalletAdd.tsx
```diff
@@ -33,8 +32,7 @@ export default function WalletAdd() {
   const navigate = useNavigate();
   const [generateMnemonic, { data: words, isLoading }] =
     useGenerateMnemonicMutation();
-  const [setLabel] = useSetLabelMutation();
-  const [addKey] = useAddKeyMutation();
+  const [addKey] = useAddPrivateKeyMutation();
```
ChiaMineJP, line 35:
```javascript
const [addPrivateKey] = useAddPrivateKeyMutation();
```
can distinct from wallet RPC API version of `addKey`

ChiaMineJP (approved):

packages/wallets/src/components/WalletAdd.tsx
paninaro: There is a subtle but important difference between:
`...(label && { label: label.trim() })`
and
`label: label.trim() ? label.trim() : undefined,`

In the first case, if label is `''` or `undefined`, it won't include the `label` field at all. In the second case, it will produce `label: undefined`

The backend behaves differently depending on the presence of `label` in the payload and will produce an error if label is an empty string or null.

Perhaps the backend should behave differently, or we could handle this at the packages/api layer, but I'm inclined to leave it as I have it now.

I'll add comments in both locations.

ChiaMineJP: OK, it makes sense.

I just thought the two are equivalent after putting values to `addPrivateKey` function.

image

image

ChiaMineJP (approved):

PR #1173, opened 2022-11-12
Offer Builder now displays totals with royalties when requesting an NFT

paninaro:
When creating an offer that requests an NFT with associated royalties, the offered fungible amounts were not showing the total amount with included royalties.

This fix will show the offered fungible amounts with royalties included.

Screenshot showing the new totals with royalties added:
image


seeden (dismissed):
lgtm

emlowe (approved):

Issue #1168, opened 2022-11-08
[BUG] Cannot read properties of undefined (reading 'map')


Taurus-gold:
### What happened?

fatal error

### Version

1.6.1

### What platform are you using?

Windows

### What ui mode are you using?

GUI

### Relevant log output or stacktrace

```shell
Cannot read properties of undefined (reading 'map')

URL
#/dashboard/nfts/nft18g6cu72sgdcrfq49dl9m8c57me4jz0xve6kjp84vpnxe5ta9f9eqp2d24w

Stacktrace
at webpack://@chia/gui/src/components/nfts/NFTMoveToProfileDialog.tsx:221:17 map
at webpack://@chia/gui/node_modules/react-dom/cjs/react-dom.production.min.js:172:114 useLayoutEffect
at webpack://@chia/gui/node_modules/react/cjs/react.production.min.js:23:112 useLayoutEffect
at webpack://@chia/gui/src/components/nfts/NFTMoveToProfileDialog.tsx:212:36 useMemo
at webpack://@chia/gui/node_modules/react-dom/cjs/react-dom.production.min.js:157:136 Ch
at webpack://@chia/gui/node_modules/react-dom/cjs/react-dom.production.min.js:267:459 b?b.sibling:null}I
at webpack://@chia/gui/node_modules/react-dom/cjs/react-dom.production.min.js:250:346 Pj
at webpack://@chia/gui/node_modules/react-dom/cjs/react-dom.production.min.js:250:277 Pj
at webpack://@chia/gui/node_modules/react-dom/cjs/react-dom.production.min.js:250:137 Pj
at webpack://@chia/gui/node_modules/react-dom/cjs/react-dom.production.min.js:243:162 Lj
```

paninaro:
Fixed in 1.7.0

PR #1175, opened 2022-11-13
Fix NFT inbox detection in NFTMoveToProfileDialog.

paninaro:
Added getNFTInbox() util function with corresponding tests.

NFTMoveToProfileDialog was using an incorrect method for identifying the NFT inbox. The incorrect code contained an unsafe dereference that could lead to an exception (caused by calling undefined.map())

`getNFTInbox` filters on NFT wallets and matches on a wallet where meta.data is falsey.

Addresses #1168


seeden (approved):
lgtm

PR #1176, opened 2022-11-14
Fix fee layout in NFT components

paninaro:
Fix EstimatedFee layout in NFT transfer/burn/did-assignment components. Without this fix, the fee component appears scrunched up.

Before:
image

Fixed:
image


seeden (approved):
lgtm

PR #1177, opened 2022-11-14
Fix/repo config

zsolt-dev:
I have separated the concerns in each commit, so please the full commit messages for my reasoning.

The airbnb rules and some additional advanced rules are setup, but there are no overrides / disables yet. I will propose those and discuss with the whole UI team.

Please do not run auto fix yet, I will run it when the rules are finalised.


seeden (approved):
lgtm

paninaro (approved):
lgtm

PR #1189, opened 2022-11-16
Store GUI preferences to `prefs.yaml`

ChiaMineJP:
## Please test the behavior
Checkout `cmj.prefs-yaml` branch and test the behavior.
The snippet:
```bash
cd chia-blockchain-gui
git fetch
git checkout cmj.prefs-yaml
npm install
npm run build
npx lerna run dev:fast –scope=@chia/gui
```

The following localStorage keys will be migrated to `prefs.yaml`:
```typescript
  const targets: string[] = [
    'cacheFolder',
    'cacheLimitSize',
    'enableAutoLogin',
    'enableDataLayerService',
    'enableFilePropagationServer',
    'darkMode',
    'finaldir',
    'fingerprintSettings',
    'hiddenWalletsItems',
    'hideObjectionableContent',
    'isHidden',
    'limit-cache-size',
    'locale',
    'mode',
    'sensitive-content',
    'skipMigration',
    'suppressShareOnCreate',
    'suppressUnsafeLinkWarning',
    'tmp2dir',
    'tmpdir',
  ];

```

The location of `prefs.yaml` is:
`$CHIA_ROOT/gui/Chia Blockchain/prefs.yaml`


seeden (changes requested):
please remove .idea directory from the PR

ChiaMineJP:
> please remove .idea directory from the PR

Done.

packages/api-react/src/hooks/usePrefs.ts
```diff
@@ -0,0 +1,56 @@
+import { useCallback, useEffect, useState } from 'react';
+import EventEmitter from '../utils/EventEmitter';
+
+const eventEmitter = new EventEmitter();
+
+function deepEqual(objA: any, objB: any){
+  return JSON.stringify(objA) === JSON.stringify(objB);
+}
+
+export default function usePrefs<T>(
```
paninaro, line 10: Can you add a type constraint on `T` to ensure that it's a JSON-compatible type? Something like what's mentioned here seems reasonable, but I'm open to other suggestions: https://dev.to/_gdelgado/type-constraints-in-typescript-34ek

packages/api-react/src/hooks/usePrefs.ts
```diff
@@ -0,0 +1,56 @@
+import { useCallback, useEffect, useState } from 'react';
+import EventEmitter from '../utils/EventEmitter';
+
+const eventEmitter = new EventEmitter();
+
+function deepEqual(objA: any, objB: any){
+  return JSON.stringify(objA) === JSON.stringify(objB);
```
paninaro, line 7: I believe that the keys in an object aren't guaranteed to be ordered in any deterministic way. So `{ foo: 1, bar: 2 }` and `{ bar: 2, foo: 1 }` may not be equal.

Looking at how this function is used (currently), I don't think it's a problem, we would just end up setting the new value even though semantically the old and new values are equivalent.

It would probably be worth adding a comment at the very least to explain that this isn't a very robust way to test for object equality so that we don't accidentally use this type of comparison elsewhere in the code.

packages/gui/src/components/plot/add/PlotAddSelectFinalDirectory.tsx
```diff
@@ -21,13 +21,14 @@ export default function PlotAddSelectFinalDirectory(props: Props) {
 
   const finalLocation = watch('finalLocation');
   const hasFinalLocation = !!finalLocation;
-  const [defaultFinalDirPath] = useLocalStorage<string>(PlotLocalStorageKeys.FINALDIR);
+  const [defaultFinalDirPath] = usePrefs<string>(PlotLocalStorageKeys.FINALDIR);
```
paninaro, line 24: Can we save `setDefaultFinalDirPath` so that we can call `setDefaultFinalDirPath(location);` instead of interacting with the window object?

```suggestion
  const [defaultFinalDirPath, setDefaultFinalDirPath] = usePrefs<string>(PlotLocalStorageKeys.FINALDIR);
```

packages/gui/src/electron/main.tsx
```diff
@@ -639,20 +652,29 @@ if (!handleSquirrelEvent()) {
...
         callback({ path: filePath });
       });
-      mainWindow?.webContents.executeJavaScript('localStorage.getItem("cacheLimitSize");', true).then((stringValue) => {
-        try {
-          cacheLimitSize = stringValue ? JSON.parse(stringValue) : cacheLimitSize;
-        } catch (e) {
+      const prefs = readPrefs();
+      if(prefs['cacheLimitSize'] !== undefined){
+        try{
+          const prefs_cacheLimitSize = +prefs['cacheLimitSize'];
```
paninaro, line 658: Is the `+` a typo?

packages/gui/src/init-prefs.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  for(const key of targets){
+    const item = window.localStorage.getItem(key);
+    if(item === undefined || item === null){
+      continue;
+    }
+
+    try{
+      prefs[key] = noParse.includes(key) ? item : JSON.parse(item);
+      window.localStorage.removeItem(key);
```
paninaro, line 55: Can we remove the items from localStorage _after_ `migratePrefs` is called successfully? If there is a problem executing migratePrefs, the user will end up with their prefs being cleared.

paninaro:
> > please remove .idea directory from the PR
> 
> Done. @seeden How about re-adding `.idea` directory to `.gitignore`? I'm using JetBrains's WebStrom and currently `.idea` directory is always shown as "Untracked files" because it was removed from `.gitignore` file.

I don't see any harm in adding `.idea` back to `.gitignore`, but we've become tightly coupled with VSCode over the past few weeks. Prettier and ESLint integrations have been setup with the VSCode project, so it's going to be an uphill battle to work in other editors going forward.

ChiaMineJP:
> > > please remove .idea directory from the PR
> > 
> > 
> > Done. @seeden How about re-adding `.idea` directory to `.gitignore`? I'm using JetBrains's WebStrom and currently `.idea` directory is always shown as "Untracked files" because it was removed from `.gitignore` file.
> 
> I don't see any harm in adding `.idea` back to `.gitignore`, but we've become tightly coupled with VSCode over the past few weeks. Prettier and ESLint integrations have been setup with the VSCode project, so it's going to be an uphill battle to work in other editors going forward.

Using VSC as a preferable IDE is OK. I think that does not block to add `.idea/` to `.gitignore` file.

packages/api-react/src/hooks/usePrefs.ts
ChiaMineJP: Makes sense.
I renamed it to `maybeEqual`.
I think it doesn't need to strictly check equality with execution cost because this check was introduced for enhancing performance.

packages/gui/src/components/plot/add/PlotAddSelectFinalDirectory.tsx
ChiaMineJP: Updated as you suggested.

packages/gui/src/electron/main.tsx
ChiaMineJP: This `+` is making sure that the value is to be a `number`.
```javascript
let str = "123";
+str === 123; // true
```

packages/gui/src/init-prefs.ts
ChiaMineJP: Thanks for checking. You're right. I updated the code as you suggested.

paninaro (approved):
lgtm

PR #1200, opened 2022-11-21
Add additional eslint airbnb rules overrides

zsolt-dev:
We are down to 1985 problems.

I have commented each override with a reason why we are overriding it.

Anyone, feel free to propose changes!


seeden (approved):
lgtm

PR #1205, opened 2022-11-22
version dialog warning

lipalong:
Added a check to compare the backend version with the gui version, and then issue a warning dialog if they're different.


packages/gui/src/components/app/AppState.tsx
```diff
@@ -148,6 +153,21 @@ export default function AppState(props: Props) {
     );
   }
 
+  if ((backendVersion && version) && (versionDialog === true)) {
+    const guiVersion = version.replace('-','.');
```
paninaro, line 157: Sorry for the back-and-forth on this. After testing this using install-gui.sh, packages/gui/package.json has version set to something like: `"1.7.0-b2.dev7"` which is transformed into `"1.7.0.b2.dev7"` by the above code. `chia version` reports `1.7.0b2.dev7`, so really we should just drop the `-`

```suggestion
    const guiVersion = version.replace('-','');
```

packages/gui/src/components/app/AppVersionWarning.tsx
```diff
@@ -0,0 +1,97 @@
...
+  open: boolean;
+  onClose: (value: string) => void;
+  backV: string;
+  guiV: string;
+}
+
+export interface WarningProps {
+  backV: string;
+  guiV: string;
+  setVersionDialog: (value: string) => void;
```
paninaro, line 25:
```suggestion
  setVersionDialog: (value: boolean) => void;
```

packages/gui/src/components/app/AppVersionWarning.tsx
```diff
@@ -0,0 +1,97 @@
...
+              The backend and the GUI are using two different versions.
+              <br />
+              <br />
+              Backend version: {backV}
+              <br />
+              GUI version: {guiV}
+              <br />
+              <br />
+              Due to this, the GUI may not behave properly.
+              </Trans>
```
paninaro, line 59:
```suggestion
              <Trans>
                The Chia application and its services are using different versions. Some functionality may not be behave
                properly as a result.
                <br />
                <br />
                Chia services version: {backV}
                <br />
                Chia application version: {guiV}
                <br />
                <br />
                It is recommended that you quit the Chia application and stop all Chia services.
              </Trans>
```

paninaro (changes requested):
Just a few small changes/suggestions and I think we're good.

paninaro (approved):

PR #1218, opened 2022-12-03
Zsolt/fix repo issues

zsolt-dev:
Please read the full commit messages for detailed info.

I would like for this to make it into the release branch, since some of the commits are affecting a ton of files


paninaro (approved):
Built and tested

PR #1238, opened 2022-12-12
Refactor all .defaultProps props to destructuring default values

zsolt-dev:
- .defaultProps for functional components is deprecated
- Default undefined props `value=undefined` were removed - redundant
- Default props for required props were removed in most cases.

### DIFF
--- a/packages/gui/src/components/offers/OfferAssetSelector.tsx
+++ b/packages/gui/src/components/offers/OfferAssetSelector.tsx
@@ -159,7 +159,6 @@ function OfferAssetSelector(props: OfferAssetSelectorProps) {
 }
 
 OfferAssetSelector.defaultProps = {
-  showAddWalletMessage: false,
   disabled: false,
 };
 
--- a/packages/gui/src/components/offers/OfferAssetSelector.tsx
+++ b/packages/gui/src/components/offers/OfferAssetSelector.tsx
@@ -102,7 +102,7 @@ type OfferAssetSelectorProps = {
 };
 
 function OfferAssetSelector(props: OfferAssetSelectorProps) {
-  const { name, id, tradeSide, defaultValue, showAddWalletMessage, onChange, ...rest } = props;
+  const { name, id, tradeSide, defaultValue, showAddWalletMessage, onChange, disabled = false, ...rest } = props;
   const { data: wallets, isLoading } = useGetWalletsQuery();
   const { data: catList = [], isLoading: isCatListLoading } = useGetCatListQuery();
   const { getValues, watch } = useFormContext();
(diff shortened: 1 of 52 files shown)


packages/gui/src/components/offers/OfferAssetSelector.tsx
```diff
@@ -159,7 +159,6 @@ function OfferAssetSelector(props: OfferAssetSelectorProps) {
 }
 
 OfferAssetSelector.defaultProps = {
```
seeden, line 161: probably you want to remove it as well

packages/gui/src/components/offers/OfferSummary.tsx
```diff
@@ -182,9 +182,3 @@ export default function OfferSummary(props: Props) {
     </Flex>
   );
 }
-
-OfferSummary.defaultProps = {
-  isMyOffer: false,
-  imported: false,
-  rowIndentation: 3,
```
seeden, line 189: I see that you just removed default props when it is defined as required in the Props.
But it is not valid as you can see for rowIndentation. 
Please change all required props to optional and use defined default values.

seeden: Please do it for all other components

seeden: rowIndentation is not used in <OfferSummary and therefore it will break functionality

seeden (changes requested):

packages/gui/src/components/offers/OfferSummary.tsx
```diff
@@ -20,12 +20,20 @@ type Props = {
...
-  rowIndentation: number;
+  rowIndentation?: number;
   setIsMissingRequestedAsset?: (isMissing: boolean) => void;
 };
 
 export default function OfferSummary(props: Props) {
-  const { isMyOffer, imported, summary, makerTitle, takerTitle, rowIndentation, setIsMissingRequestedAsset } = props;
+  const {
+    isMyOffer,
+    imported,
```
seeden, line 30: missing default value

packages/gui/src/components/offers/OfferSummaryRow.tsx
```diff
@@ -43,10 +43,6 @@ function OfferMojoAmount(props: OfferMojoAmountProps): React.ReactElement | null
   );
 }
 
-OfferMojoAmount.defaultProps = {
-  mojos: 0,
```
seeden, line 47: missing default value

seeden (changes requested):
many default values are missing in components. please check whole PR again

seeden (approved):
lgtm

PR #1243, opened 2022-12-13
Enforce the style of numeric separators by correctly grouping digits

zsolt-dev:
This PR contains an eslint rule change and also changes in the code, so it follows the suggested rule.

We use a lot of big numbers (Mojos etc), so I believe this is necessary
If there was a comment after the a big number, the comment was removed


seeden (dismissed):
lgtm. I will wait for previous PR and merge it after that

PR #1246, opened 2022-12-15
Enforce the style of numeric separators by correctly grouping digits

zsolt-dev:
We use a lot of big numbers (Mojos etc), so I believe this is necessary
If there was a comment after the a big number, the comment was removed


seeden (approved):
lgtm

PR #1249, opened 2022-12-15
Add check commands and refactor prettier config

zsolt-dev:
- new script: npm run check which will run TS, Eslint and prettier checks
- npm run check:types will run TS check in each repo
- prettier config updated to prettify non js files too
- lint-staged config removed, since it is not used at all (lint-staged not installed)


paninaro (approved):
lgtm

PR #1257, opened 2022-12-17
Refactor NFT metadata LRU implementation

paninaro:
Adds an LRU cache class and factory function in @chia-network/core.

Adds a React Context named LRUsContext that maintains a mapping of named LRUs. Currently we only use an LRU named "nftMetadata" for managing the in-memory NFT metadata cache. The LRUsContext provider is installed in AppProviders.

Updates the previous LRU call sites to use the new `useNFTMetadataLRU` hook, providing an object oriented interface to using the LRU.

Tests added to fully cover the LRU class implementation.


paninaro:
@seeden Is it appropriate to install the new context in AppProviders? My goal is to have the LRU cache associated with the currently selected wallet key.

rifeljm (approved):

PR #1261, opened 2022-12-22
Fixed an issue where text field component did not recognize rules

ChiaMineJP:


packages/core/src/components/TextField/TextField.tsx
```diff
@@ -49,39 +48,100 @@ export type TextFieldProps = MaterialTextFieldProps & {
...
+ * Convert original form rules to `react-hook-form` compatible rules.
+ */
+function normalizeRules<T>(rules?: ReactRules<T>): ControllerProps<ReactElement<T>>['rules']|undefined {
+  if(!rules){
+    return undefined;
+  }
+  
+  const newRules: ControllerProps<ReactElement<T>>['rules'] = {};
+  
+  if(typeof rules.min === 'object' && rules.min.value !== undefined){
```
paninaro, line 61: nit: I'm guessing that prettier hasn't been used to format these changes. Can you re-save the files in VSCode with the latest ESLint/prettier extensions installed?

ChiaMineJP: So VSCode is preferred for developing the GUI...
I personally love JetBrains' WebStorm for Typescript/Javascript developing but I'll do with VSCode from now on.

ChiaMineJP:
There are lint errors but they are from files which this PR does not involve.
I think this PR shouldn't fix the lint errors because it may introduce conflicts.

PR #1262, opened 2022-12-22
Nft gallery optimizations (support for loading a few thousand NFTs)

rifeljm:


packages/gui/src/components/offers2/OfferBuilderNFT.tsx
```diff
@@ -106,9 +106,9 @@ export default function OfferBuilderNFT(props: OfferBuilderNFTProps) {
                   nft={nft}
                   canExpandDetails={false}
                   availableActions={
-                    NFTContextualActionTypes.CopyNFTId |
-                    NFTContextualActionTypes.ViewOnExplorer |
-                    NFTContextualActionTypes.OpenInBrowser |
+                    NFTContextualActionTypes.CopyNFTId ||
+                    NFTContextualActionTypes.ViewOnExplorer ||
+                    NFTContextualActionTypes.OpenInBrowser ||
```
paninaro, line 111: We definitely don't want to use logical OR here, but since ESLint doesn't like the use of bitwise OR, we can just add the values as:

```suggestion
                    NFTContextualActionTypes.CopyNFTId +
                    NFTContextualActionTypes.ViewOnExplorer +
                    NFTContextualActionTypes.OpenInBrowser +
```

paninaro: Resolved

paninaro (approved):

PR #1274, opened 2023-01-05
moving left and right between nft details

rifeljm:


packages/gui/src/components/nfts/gallery/NFTGallery.tsx
```diff
@@ -27,6 +27,8 @@ import SelectedActionsDialog from './SelectedActionsDialog';
 
 export const defaultCacheSizeLimit = 1024; /* MB */
 
+export const shownNFTs = new Map();
```
paninaro, line 30: We shouldn't export or provide direct access to this object. I would probably use a Context/Provider/hook to wrap this up nicely. You can take a look at LRUsProvider/useNFTMetadataLRU as a possible example. Or I'm sure @seeden would have some excellent advice as well.

paninaro (changes requested):
Seth has an updated design. Can you match what he has in Figma? https://www.figma.com/file/uCI6CTQ1W4uElMtmN6P4Qv/H2-2022?node-id=2256%3A89973&t=IenVQXXjYQSp5atr-0

PR #1277, opened 2023-01-05
detail back to gallery - scroll to same position

rifeljm:


packages/gui/src/components/nfts/gallery/NFTGallery.tsx
```diff
@@ -28,6 +28,8 @@ export const defaultCacheSizeLimit = 1024; /* MB */
 
 const maxNFTsPerPage = 200;
 
+const scrollPosition = new Map();
```
paninaro, line 31: No need for this to be a map.

```suggestion
let scrollPosition = 0;
```

packages/gui/src/components/nfts/gallery/NFTGallery.tsx
```diff
@@ -253,6 +254,21 @@ export default function NFTGallery() {
...
+      if (document.getElementById('scroll-helper') && scrollPosition.get('position') > 0) {
+        setTimeout(() => {
+          if (document.getElementById('scroll-helper')?.parentNode) {
+            (document.getElementById('scroll-helper')?.parentNode as HTMLElement).scrollTo(
+              0,
+              scrollPosition.get('position')
+            );
+          }
+        }, 0);
+      }
```
paninaro, line 268: A couple of suggestions:

1) `document.getElementById('scroll-helper')` is called up to 3 times. Can we just assign the result to a local const? ```const scrollHelper = document.getElementById('scroll-helper');```

2) I'm finding it harder and harder to read the code with so many useEffects and other hooks that have inlined functions. To improve the code clarity, I would like to see a couple of functions named `setScrollPosition()` and `restoreScrollPosition()`. Can we refactor this code to support that?

packages/gui/src/components/nfts/gallery/NFTGallery.tsx
```diff
@@ -409,9 +425,9 @@ export default function NFTGallery() {
 
   return (
     <LayoutDashboardSub
-      ref={dashboardSub}
       // sidebar={<NFTGallerySidebar onWalletChange={setWalletId} />}
       onScroll={(e: MouseEvent) => {
+        scrollPosition.set('position', (e.target as HTMLElement).scrollTop);
```
paninaro, line 430: From the above comment:
`setScrollPosition((e.target as HTMLElement).scrollTop);`

packages/gui/src/components/nfts/gallery/NFTGallery.tsx
```diff
@@ -253,6 +254,21 @@ export default function NFTGallery() {
     };
   }, []);
 
+  React.useEffect(() => {
+    if (allowNFTsFiltered.length) {
+      if (document.getElementById('scroll-helper') && scrollPosition.get('position') > 0) {
+        setTimeout(() => {
+          if (document.getElementById('scroll-helper')?.parentNode) {
+            (document.getElementById('scroll-helper')?.parentNode as HTMLElement).scrollTo(
```
paninaro, line 262: Currently this is setting the scroll position whenever the NFTGallery is mounted. I was hoping that the scroll position would only be restored when navigating back from NFTDetails.

paninaro (changes requested):
A few things to address

packages/gui/src/components/nfts/gallery/NFTGallery.tsx
paninaro: Another problem with this is that the scroll position is remembered across all wallet keys. So setting the scroll position in a gallery with hundreds of NFTs, and then switching to another wallet key, the NFT gallery would try to restore the scroll position from the prior wallet key's gallery.

PR #1281, opened 2023-01-06
Update version check to account for '+' (PEP440 local version)

paninaro:
Fix version comparison issue with custom builds that make use of the PEP440 local version component. Reported by the community.

Testing notes:
Tested with a backend and frontend version string that contained `+og.1.2.0` and `-og.1.2.0` respectively. Also tested expected version strings for normal builds.


lipalong (approved):
Looks good.

PR #1284, opened 2023-01-07
Add 12-word mnemonic import option

lipalong:
Added the option to import from a 12-word mnemonic in addition to the existing functionality of importing from a 24-word mnemonic.


paninaro (approved):
lgtm. Tested with 12 and 24 word mnemonics.

PR #1290, opened 2023-01-09
Refactor all forms that are not using react-hook-form

zsolt-dev:
- Tested (including submitting via pressing enter key)
- react-hook-form had to be updated to support `shouldSelect: true`
- Fixed bug: ChangePassphrasePrompt was never displayed


paninaro (approved):
Did some testing. LGTM.

PR #1306, opened 2023-01-13
Nft gallery multiple selection move to profile

rifeljm:


packages/api/src/wallets/NFT.ts
```diff
@@ -33,13 +33,25 @@ export default class NFTWallet extends Wallet {
...
+    return Promise.all(
+      nftCoinIds.map((nftCoinId) =>
+        this.command('nft_set_nft_did', {
+          walletId,
+          nftCoinId,
+          didId: did,
+          fee,
+        })
+      )
+    ).then((responses) => responses);
```
paninaro, line 54: I may have miscommunicated the RPC changes we need to support bulk NFT DID assignment. We should be calling `nft_set_nft_did` for assigning a DID to a single NFT, but for assigning a DID to _multiple_ NFTs, we should instead be calling the new RPC `nft_set_did_bulk`

So we should have a new function like:
```
async setNftDidBulk(nftCoinList: string[], fee: string, didId: string | undefined) {
    ...
    this.command('nft_set_did_bulk',  {
        nftCoinList,
        didId,
        fee,
    })
    ...
```

The RPC parameters are named `nftCoinList`, `didId`, and `fee`

It's valid for `didId` to be undefined, in which case, the DID will be removed from the specified NFTs.

packages/api-react/src/services/wallet.ts
```diff
@@ -2025,15 +2025,15 @@ export const walletApi = apiWithTag.injectEndpoints({
       {
         walletId: number;
         nftLauncherId: string;
-        nftCoinId: string;
+        nftCoinIds: string[];
```
paninaro, line 2028: Please see my comment below. We should add another API here: `setNFTDIDBulk` that will call `setNftDidBulk()`

paninaro (changes requested):
The UI changes look good to me. We just need to make the RPC changes described in the comments.

paninaro (approved):

paninaro (approved):

PR #1331, opened 2023-01-19
limit number of nfts = 10

seeden:
added limit for number of nfts in offer builder


paninaro (commented):
Thanks for making this change. Would you mind also adding the limit in `NFTCreateOfferContextualAction` to handle the multiple selection case from the NFT gallery?

paninaro (approved):

PR #1339, opened 2023-01-20
Fix some TS issues

zsolt-dev:
Please read the individual commit messages for more info


packages/icons/package.json
```diff
@@ -43,6 +43,7 @@
     "@types/react": "17.0.38",
     "@types/react-dom": "17.0.11",
     "@types/react-router-dom": "5.3.2",
+    "@types/styled-components": "^5.1.26",
```
paninaro, line 46: Would you be ok with pinning to an exact version?

paninaro (approved):

PR #1349, opened 2023-01-24
Update offer builder to not require any assets on the requesting side…

rifeljm:
… when creating an offer


paninaro (changes requested):
Looks like the code needs reformatting. The code check CI task is failing:
image

paninaro (approved):

PR #1357, opened 2023-01-25
Fix remaining TS in icons package and add to CI

zsolt-dev:
importing icons in other packages now correctly exposes types.


packages/icons/src/NFTs.tsx
```diff
@@ -19,5 +19,5 @@ export function Reload(props: SvgIconProps) {
 }
 
 export function Copy(props: SvgIconProps) {
-  return <CopyIcon component={ReloadIcon} viewBox="0 0 22 22" {...props} />;
+  return <SvgIcon component={CopyIcon} viewBox="0 0 22 22" {...props} />;
```
paninaro, line 22: Looks like the prior code was a hacky change to update all of the CopyToClipboard icons.

With this change, all of the Copy icons are incorrect:

image

paninaro (changes requested):
Looks good. Just need a change for the Copy icon.

paninaro (approved):

paninaro (approved):

PR #1363, opened 2023-01-26
Rifeljm/#203 multiple selection burn

rifeljm:


paninaro (approved):
I see one issue with this. If the selected NFTs are from different wallet ids, I believe the transfer will fail. I'll double check that and file an issue if that's the case.

Approving as-is

paninaro:
Tested burning from different wallet ids and everything worked as expected

PR #1381, opened 2023-01-29
fix enable auto login

### LINKED ISSUE #1380: [Bug] Enable Auto Login isn't working
When Enable Auto Login is toggled on, launching the GUI doesn't skip the key selection screen.

There may have been a subtle change in how the wallet node starts up that could impact this feature. Needs investigation.

paninaro:
also removes the "Delete All Keys" option when no keys are present

Fixes #1380


seeden (dismissed):
lgtm

seeden (approved):
lgtm

PR #1401, opened 2023-02-01
fix wording in offer sharing summary dialog (fix #1399)

### LINKED ISSUE #1399: Share Offer panel conflates Maker and Taker [Bug]
### What happened?

To reproduce:
1. Maker uses the reference wallet GUI to create an offer to buy an NFT in exchange for XCH.
2. Share Offer panel appears. Maker clicks `DEXIE`.
3. The wording on the panel is "In exchange for: [number] XCH" and "Your offer: [NFT details]"

This is shown from the point of view of the Taker, but the user here is the Maker. Either the assets should be swapped, or the wording should be changed. For consistency, I think it makes more sense to stick with the same wording as the Offer Builder. The `DEXIE` panel could say `Offering` and `Asset I would like to trade` followed by the amount of XCH, and then `Requesting` and `Assets I would like to receive` followed by the NFT details.

### Version

1.6.2, but this issue has existed for a long time

### What platform are you using?

Windows

### What ui mode are you using?

GUI

### Relevant log output or stacktrace

_No response_

paninaro:


packages/gui/src/components/offers/OfferShareDialog.tsx
```diff
@@ -1481,7 +1481,7 @@ export default function OfferShareDialog(props: OfferShareDialogProps) {
       </DialogTitle>
 
       <DialogContent dividers>
-        <Flex flexDirection="column" gap={2}>
+        <Flex flexDirection="column" gap={2} paddingTop="16px">
```
seeden, line 1484: Can you use the number 2 instead of the string 16px? material-ui will replace it with a correct value, which is always multiplied by 8px

1 = 8px
1.5 = 12px
2 = 16px
...

seeden (changes requested):

PR #1404, opened 2023-02-01
reject object attribute values when rendering NFT details

paninaro:
fixes exceptions raised when dealing with unsupported/malformed NFT attributes/rankings


seeden (approved):
lgtm

PR #1411, opened 2023-02-02
Notification icon redesign (#1407)

paninaro:
Cherry-picked from main commit c107aec271714a0666836c1bce86083a7f344916


packages/gui/src/components/notification/NotificationsDropdown.tsx
```diff
@@ -33,7 +33,7 @@ export default function NotificationsDropdown() {
           sx={buttonStyle}
         >
           <Badge color="primary" badgeContent={unseenCount} invisible={!unseenCount}>
-            <NotificationsNoneIcon color="secondary" />
+            <NotificationIcon color="secondary" />
```
rifeljm, line 36: secondary color inside the theme means "#000", but Figma has #757575 for basically all the icons in top section of the page

rifeljm: also svg inherits color="secondary", so if we remove color="secondary" only in the icon, it will still stay black :/

rifeljm (approved):

packages/gui/src/components/notification/NotificationsDropdown.tsx
paninaro: I'll create an issue to track updating the theme with a "headerIcon" type or similar.

PR #1415, opened 2023-02-02
fix crash when creating a plotNFT (fix 1396)

paninaro:
fixes a bad dereference on a hook dependency array addition.


seeden (approved):
lgtm

PR #1417, opened 2023-02-02
launch daemon if self_hostname === ui.daemon_host

paninaro:
When overriding config.yaml values that specify localhost by default, this change will allow the GUI to launch the daemon if `self_hostname` equals `ui:daemon_host`

Fixes issue https://github.com/Chia-Network/chia-blockchain/issues/14468

Repro steps for the original issue:

1) Update config.yaml such that all localhost values are replaced by the machine's local IP (e.g. 192.168.1.100)
2) With the daemon and all services stopped, launch the GUI

Observe:

1) The GUI fails to launch the daemon and is stuck on the loading keyring screen


seeden (approved):
lgtm

PR #1426, opened 2023-02-03
Refactor build scripts for faster dev and CI builds

zsolt-dev:
- new command: `npm run build:skipLocales` -> faster build time by 30% and does not generate locales
- new command: `npm run build:ci` -> faster build time by skipping locale:extract (only running locale:compile) - consulted with Zlatko
- renamed command `npm run dev:fast` to `npm run dev:skipLocales`


paninaro (approved):
🎉

PR #1451, opened 2023-02-08
Remove fees from add cat token

Ganbin:
Fees are not needed to add or rename a CAT token.


paninaro (approved):
Thanks for the contribution. Looks good to me.

PR #1454, opened 2023-02-09
Show memo when expanding inbound transaction

rifeljm:


packages/wallets/src/components/WalletHistory.tsx
```diff
@@ -197,17 +202,26 @@ export default function WalletHistory(props: Props) {
...
     rowsPerPage,
     count,
     pageChange,
   } = useWalletTransactions(walletId, 10, 0, 'RELEVANCE');
+
+  React.useEffect(() => {
+    if (transactionWithoutIncomingMemos?.length) {
+      setTransactions(transactionWithoutIncomingMemos);
+    }
+  }, [transactionWithoutIncomingMemos]);
```
paninaro, line 219: Is this effect necessary? It's not obvious to me why we need this.

packages/wallets/src/components/WalletHistory.tsx
```diff
@@ -249,6 +263,17 @@ export default function WalletHistory(props: Props) {
         metadata={metadata}
         expandedCellShift={1}
         uniqueField="name"
+        onToggleExpand={async (rowId: string, expanded: boolean) => {
+          const memos = (await getTransactionMemo({ transactionId: rowId.split('-')[0] })).data;
```
paninaro, line 267: We only want to call this if expanded, right? Otherwise it gets called when expanding or collapsing.

packages/wallets/src/components/WalletHistory.tsx
```diff
@@ -249,6 +263,17 @@ export default function WalletHistory(props: Props) {
...
+          if (expanded) {
+            const newTransactions = transactions.map((transaction) =>
+              transaction.name === rowId.split('-')[0] && memos && Object.keys(memos).length > 0
+                ? { ...transaction, memos }
+                : transaction
+            );
+            setTransactions(newTransactions);
```
paninaro, line 274: This seems really inefficient. Instead of replacing the entire collection of transactions with a modified version, what if we just maintained a map of transaction IDs and their memos (previously fetched). Then when we look for the memo on a row, try to pull it from the row first, and if not there, try to pull it from the map.

rifeljm: This is me being lazy :)
Fixed.

packages/wallets/src/components/WalletHistory.tsx
rifeljm: You're right. Fixed.

packages/wallets/src/components/WalletHistory.tsx
rifeljm: Well, you might be right on this one, but I just didn't and still don't know how to put memo inside a rendered row other way.

The problem is that we need to mutate "transactions" inside "TableControlled" component each time we fetch a new memo and it is the "cols" variable that defines how each row will be rendered and the rendering rule is defined outside the the main component, so we can not access state from each row.

```
      <TableControlled
        cols={cols}
        rows={transactions ?? []}
```        

I was also tempted to create another hook for that in a separate file, but I would then need to put useEffect there, so I just did it this way with less code.

packages/wallets/src/components/WalletHistory.tsx
```diff
@@ -192,6 +192,8 @@ type Props = {
   walletId: number;
 };
 
+const mapMemos: any = {};
```
paninaro, line 195: I would have used `useState({})` to avoid creating a global variable (bad)

packages/wallets/src/components/WalletHistory.tsx
```diff
@@ -263,9 +265,10 @@ export default function WalletHistory(props: Props) {
...
         expandedCellShift={1}
         uniqueField="name"
-        onToggleExpand={async (rowId: string, expanded: boolean) => {
-          const memos = (await getTransactionMemo({ transactionId: rowId.split('-')[0] })).data;
-          if (expanded) {
+        onToggleExpand={async (rowId: string, wasExpanded: boolean) => {
+          if (wasExpanded) {
+            const memos = mapMemos[rowId] || (await getTransactionMemo({ transactionId: rowId.split('-')[0] })).data;
+            mapMemos[rowId] = memos;
             const newTransactions = transactions.map((transaction) =>
```
paninaro, line 272: Looks much better, thanks. If I'm being nitpicky, I would only change mapMemos[rowId] IF the memo was just fetched from getTransactionMemo. In other words, there's no need to update the dict if the memo came from there.

packages/wallets/src/components/WalletHistory.tsx
paninaro: I looked into using the LRU code, but since this component is in the wallets package, it would've involved moving a bunch of code. I'm ok with this being a global for now.

packages/wallets/src/components/WalletHistory.tsx
paninaro: I made a few additional changes in this area. Resolving this now.

paninaro (approved):

PR #1457, opened 2023-02-10
Absorb failure when unable to determine remote content length

paninaro:
When fetching NFT content/metadata, we attempt to fetch the content length using a HEAD request, and then proceed with downloading the content. If the HEAD request fails for some reason, an exception was being thrown which would prevent the content from loading.

On Windows, a strange set of events could cause the user to see `EPERM: operation not permitted, open '<path>'` error dialogs. The sequence of events seems to be:

1) fetchBinaryContent opens a file stream for writing
2) Performing a HEAD request to fetch the content length would fail, raising an exception
3) The promise is rejected

meanwhile...

4) A separate call to fetchBinaryContent proceeds
5) Performing the HEAD request succeeds
6) The data is downloaded
7) When determining whether to cache the data, the cache directory is probed
8) While iterating over the cache directory, the 0-byte file from the previously failed call can't be stat'd leading to the EPERM error.

For this patch, the (lack of) error handling in `getCacheSize()` hasn't been addressed, but rather the file stream is now closed and the call to `getRemoteFileSize()` is placed in a try/catch, allowing the request to be handled without learning the content length upfront.


paninaro:
Superseded by #1467

PR #1460, opened 2023-02-10
fix useLocalStorage's default value behavior when a callback is used in the setter function

paninaro:
Using code like the below:

```
const [something, setSomething] = useLocalStorage('something', {})
...
setSomething((currentSomething) => ...);
```
Within the callback provided to setSomething, currentSomething was set to `undefined` if the local storage key `'something'` didn't already exist. This is contrary to the expected behavior where currentSomething should be `{}`, as specified as the default value.

This change will substitute the `undefined` value with the default value at most one time. We only substitute the default value once so that it's possible to explicitly set the value back to `undefined`


paninaro:
Closing as https://github.com/Chia-Network/chia-blockchain-gui/pull/1470 has been merged

PR #1476, opened 2023-02-16
fix nft royalties bug on create offer page

lipalong:
Fix bug on create offer page where NFT royalties continued to be shown after requested NFT was removed


paninaro (approved):
lgtm

PR #1477, opened 2023-02-17
Wallet transaction list UX tweaks

rifeljm:
- truncated wallet address
- moved the wallet address column to the far right
- expand/collapse button will only be hidden if screen width will be less than 1150px

### DIFF
--- a/packages/wallets/src/components/WalletHistory.tsx
+++ b/packages/wallets/src/components/WalletHistory.tsx
@@ -182,7 +182,7 @@ const getCols = (type: WalletType, isSyncing, getOfferRecord, navigate) => [
   },
 
   {
-    width: '60px',
+    width: '66px',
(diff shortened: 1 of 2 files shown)


packages/wallets/src/components/WalletHistory.tsx
```diff
@@ -141,45 +180,9 @@ const getCols = (type: WalletType, isSyncing, getOfferRecord, navigate) => [
...
-          <FormatLargeNumber value={mojoToChia(row.feeAmount)} />
-        </strong>
-        &nbsp;
-        {metadata.feeUnit}
-      </>
-    ),
-    title: <Trans>Fee</Trans>,
-  },
   {
+    width: '60px',
```
paninaro, line 185: When hovering over the expand/collapse button, the highlighted circle is cut off

```suggestion
    width: '66px',
```

paninaro (approved):

PR #1479, opened 2023-02-17
display CAT amount when using spendCAT via WalletConnect (fixes #1254)

### LINKED ISSUE #1254: WalletConnect spendCAT confirmation shows XCH amount
When the spendCAT WalletConnect command displays its confirmation, the amount is shown as XCH instead of the CAT type.

paninaro:
Changes:
Adds WalletConnectCommandParamName enum to replace WalletConnectCommandParam's name type
Adds WalletConnectCATAmount component to render CAT mojos
useWalletConnectCommand's displayComponent function now passes the full list of params when rendering the param value


seeden (approved):
lgtm

PR #1498, opened 2023-02-27
expand offered fee section by default for imported offers

paninaro:
When viewing an offer that was imported (from an offer file or data), the offered fee section will now be expanded by default. Behavior is unchanged when creating a new offer or viewing an offer that has already been accepted.

image


seeden (approved):
lgtm

PR #1508, opened 2023-03-01
wallet connect improvements

seeden:
added support for bypass commands and remember choice
added did commands and missing commands for wallet connect 
added missing properties
added offer viewer during takeOffer command


packages/gui/src/constants/WalletConnectCommands.tsx
```diff
@@ -462,6 +526,7 @@ const walletConnectCommands: WalletConnectCommand[] = [
     command: 'transferNFT',
     label: <Trans>Transfer NFT</Trans>,
     service: ServiceName.WALLET,
+    bypassConfirm: true,
```
paninaro, line 529: I think including this as a command that doesn't always require confirmation was probably a mistake. I'm checking with E-ming.

paninaro:
This is looking great. I did notice that when I have multiple Dapps connected, remembering my choice for a command like getSyncStatus ends up saving that pref for all Dapps instead of just for the requesting Dapp.

packages/gui/src/constants/WalletConnectCommands.tsx
seeden: Updated

paninaro (approved):

PR #1521, opened 2023-03-04
show NFT title and preview in offer rows

paninaro:
Display NFT title and preview in offer rows when the offered/requested sections contain a single NFT asset. For mixed assets on one side of the offer, we'll display the NFT title but not the preview. A future redesign will be more comprehensive.

Examples:
image
image
image


seeden (approved):
lgtm

PR #1527, opened 2023-03-06
add default fee to fee estimator

lipalong:
Added a default value to the fee estimator. The default value is selected to be the 0-valued fee from the available fee options. If such an option does not exist, a 0-valued fee option is created and added to the selectable list of fee estimates.


packages/core/src/components/EstimatedFee/EstimatedFee.tsx
```diff
@@ -261,7 +289,7 @@ export default function EstimatedFee(props: FeeProps) {
         }
       }
     }
-  }, [formattedEstimates, name, selectedTime, setValue]);
+  }, [formattedEstimates, name, selectedTime, setValue, defaultFee]);
```
paninaro, line 292:
```suggestion
  }, [formattedEstimates, name, selectedTime, setValue, defaultFee, isLoading]);
```

paninaro (approved):

PR #1531, opened 2023-03-07
Randomizing the color and emoji for existing keys the first time use …

rifeljm:
…runs 1.8.0


paninaro (changes requested):
Looks good. Let's update this PR to select emojis from the nature/animals and smileys sets.

paninaro (approved):

PR #1544, opened 2023-03-08
disabled counter offer button tooltip

lipalong:
Add tooltip to disabled Counter button on Offers page to inform user that the sender chose not to allow counter offers


paninaro (approved):
lgtm

PR #1548, opened 2023-03-09
fix countering offer flow and allow countering from viewing an offer from a notification

paninaro:
The core change here is that there's a check that's performed to determine if the NFTs in an offer belong to the user. The result of this check informs how the send notification dialog behaves -- whether to show the NFT details for an NFT the user is attempting to acquire.


seeden (approved):
lgtm

PR #1549, opened 2023-03-09
added support for reset bypass

seeden:
added support for reset bypass


packages/gui/src/components/settings/SettingsIntegration.tsx
```diff
@@ -82,6 +85,38 @@ export default function SettingsIntegration() {
...
+                <Trans>Reset</Trans>
+              </Button>
+            </Flex>
+          </Flex>
+        </Grid>
+        <Grid item container style={{ width: '400px' }} gap={2}>
+          <SettingsText>
+            <Trans>
+              Take control of your WalletConnect security with reset bypass. Click the button to revoke previous command
+              approvals and ensure all future commands require your approval.
```
paninaro, line 115:
```suggestion
This will reset all previously granted permissions across all Dapps that have been connected to. After
              resetting you will be asked to grant permission again.
```

paninaro (approved):

PR #1555, opened 2023-03-10
Emoji and color picker positioning fix (did not work at the bottom)

rifeljm:


paninaro (approved):
Increasing the bottom padding is fine as a workaround for 1.7.1. I'd prefer a solution that behaves like a popover at some point, but we can revisit it later.

PR #1561, opened 2023-03-11
added walletconnect app permissions gui

Gaerax:
The new section allows users to manage the permissions that they have granted to each app.
Users can choose to automatically confirm or reject any read-only commands by toggling the switch, or delete the permission to be prompted every time the command is run.

image


paninaro (approved):
There are a few styling changes that we should make, but this looks good. Merging this into another branch so that we can incorporate a few more changes.

PR #1575, opened 2023-03-14
bumping playwright

wallentx:
They had some good changes


paninaro (approved):

paninaro (changes requested):
Need to update to make these dev/optional dependencies

PR #1587, opened 2023-03-15
Rifeljm/#1528 theme picker same color

rifeljm:


paninaro (changes requested):
This PR affects the border color of other components like WalletReceiveAddressField. I don't disagree with the approach here, but for this release I'd like us to follow the existing pattern for setting the border color.

In EmojiAndColorPicker, we can use:

```
border: `1px solid ${isDark ? theme.palette.border.dark : theme.palette.border.main}`,
```

paninaro (approved):

PR #1595, opened 2023-03-15
spend_type param

lipalong:
Update the Fee Estimator's get_fee_estimate calls to use the spend_type parameter


paninaro (approved):
Looks good. Thanks!

PR #1598, opened 2023-03-15
Seeden/unknown cat amount fix2

seeden:
fixed unknown CAT amount
fixed XCH showed as unknown CAT
fixed seenHeight


paninaro (approved):
lgtm. Works properly in my testing.

PR #1607, opened 2023-03-16
Check latest GUI version on http://chia.net/downloads

rifeljm:
and show reminder dialog if there is a newer version of GUI


paninaro:
Moved this to a draft as there are other considerations to factor in.

paninaro (approved):

Issue #1618, opened 2023-03-17
[Bug] limiting the number of rows to show in the conflicting offers screen doesn't work


esaung:
### What happened?

when shown conflicting offers the window shows all conflicting offers and doesn't limit the number of rows based on the option selected. Changing the selected value doesn't change the number of rows shown. We don't need the row filter for this screen.

![Screenshot 2023-03-17 at 12 06 09 PM](https://user-images.githubusercontent.com/93196638/225975501-59d31815-823d-4181-a869-510dcf646671.png)

### Version

1.7.1 rc2

### What platform are you using?

macOS

### What ui mode are you using?

GUI

PR #1625, opened 2023-03-18
[BUGFIX] limiting the number of rows to show in the conflicting offers screen doesn't work

ChiaMineJP:
This PR fixes https://github.com/Chia-Network/chia-blockchain-gui/issues/1618


paninaro (approved):
Tested and found to be working as expected.

PR #1645, opened 2023-03-22
Refactor styled components to sx prop in core package

zsolt-dev:


paninaro:
The wallet key home screen has some rendering issues after applying this set of changes:

image
image

paninaro (changes requested):
@zsolt-dev I did some additional testing and everything seems to be working well with the exception of the wallet key home screen. Please see the screenshots in my early comment.

packages/core/src/components/CardListItem/CardListItem.tsx
```diff
@@ -45,14 +16,31 @@ export type CardListItemProps = {
...
   );
 
   return (
-    <StyledCard variant="outlined" selected={selected} disabled={disabled} {...rest}>
+    <Card
+      variant="outlined"
+      {...rest}
+      sx={{
+        width: '100%',
+        borderRadius: 1,
```
paninaro, line 31: In my testing, `1` doesn't set the correct border radius. Setting it as below works as expected.

```suggestion
        borderRadius: (theme) => `${theme.spacing(1)}`,
```

packages/core/src/components/CardListItem/CardListItem.tsx
```diff
@@ -45,14 +16,31 @@ export type CardListItemProps = {
 export default function CardListItem(props: CardListItemProps) {
   const { children, selected, onSelect, loading, disabled, noPadding = false, ...rest } = props;
 
-  const content = noPadding ? (
-    <StyledCardContentNoPadding>{children}</StyledCardContentNoPadding>
-  ) : (
-    <StyledCardContent>{children}</StyledCardContent>
+  const content = (
+    <CardContent sx={{ marginBottom: (theme) => (noPadding ? 0 : `${theme.spacing(2)} !important`) }}>
+      {children}
```
paninaro, line 21: Previously, `StyledCardContent` and `StyledCardContentNoPadding` set padding/paddingBottom, not marginBottom. 

This is affecting the layout of the wallet keys, and may also affect other areas of the UI (though I'm unsure where).

The padding should be 0 in this case:

Wrong:
image

Correct:
image

paninaro (approved):

PR #1648, opened 2023-03-27
Add absent libgbm1 dependency

Elentary:
Fixes [issue 14554 from the main repo](https://github.com/Chia-Network/chia-blockchain/issues/14554)

`dpkg -I` applied to the latest .deb of chia-blockchain GUI shows the following list of dependencies:

`Depends: libgtk-3-0, libnotify4, libnss3, libxss1, libxtst6, xdg-utils, libatspi2.0-0, libuuid1, libsecret-1-0`

However, as it was mentioned in the [issue 14554](https://github.com/Chia-Network/chia-blockchain/issues/14554) from the main chia-blockchain repo, this .deb lacks the dependency from `libgbm1`, and if being run shows the following error:

```
/# chia-blockchain
chia-blockchain: error while loading shared libraries: libgbm.so.1: cannot open shared object file: No such file or directory
```

So I added the missing dependency in `package.json` according to [the electron-builder docs](https://www.electron.build/configuration/linux#debian-package-options). I have checked with`dpkg -I` that .deb package built from my PR had this required dependency and no existed dependency had been lost:

`Depends: libgbm1, libgtk-3-0, libnotify4, libnss3, libxss1, libxtst6, xdg-utils, libatspi2.0-0, libuuid1, libsecret-1-0`

Also I had to explicitly list all default dependencies in the `package.json` in order not to override them. For clarity I made it in a separate commit.

### DIFF
--- a/packages/gui/package.json
+++ b/packages/gui/package.json
@@ -93,6 +93,7 @@
       "afterInstall": "../../../build_scripts/assets/deb/postinst.sh",
       "afterRemove": "../../../build_scripts/assets/deb/prerm.sh",
       "depends": [
+        "libgbm1",
         "libgtk-3-0", "libnotify4", "libnss3", "libxss1", "libxtst6", "xdg-utils", "libatspi2.0-0", "libuuid1", "libsecret-1-0"
       ]
     },
--- a/packages/gui/package.json
+++ b/packages/gui/package.json
@@ -91,7 +91,10 @@
     },
     "deb": {
       "afterInstall": "../../../build_scripts/assets/deb/postinst.sh",
-      "afterRemove": "../../../build_scripts/assets/deb/prerm.sh"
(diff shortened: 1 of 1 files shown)


paninaro:
@ChiaMineJP Would you mind taking a look?

ChiaMineJP (approved):
LGTM.
Thanks for the PR!

If you can afford, could you tell us your Ubuntu environment? Because `libgbm1` seems to be installed by default when you setup Ubuntu Desktop 22.04 from `.iso`

Elentary:
I wish I could.
The original issue author mentioned `Debian 11.1 based Devuan, amd64`.
I was able to reproduce it in docker, so can't say whether it works with vanilla Ubuntu Desktop.
Generally, I believe we still have to specify all the deps, coz some day Ubuntu Desktop default environment could be easily changed without notice to chia-devs

PR #1649, opened 2023-03-28
Add react and redux devtools to electron

zsolt-dev:
The electron-devtools-assembler is a new NPM package with only a few installs. It is a fork of popular package and I have checked the whole diff between them and locked the version, so its all good.


paninaro (approved):
Amazing 🥳

PR #1651, opened 2023-03-29
NFT Provider - refactored NFTs

seeden:
refactored nfts
used one cache instead of many
simplified code


packages/api-react/src/hooks/useGetNFTsCount
paninaro, line 1: Remove?

seeden: removed

paninaro (approved):

PR #1652, opened 2023-03-29
fix acceptance of one-sided offers

paninaro:
`offerBuildDataToOffer()` now takes an optional boolean parameter that indicates whether the offer column can be empty

### DIFF
--- a/packages/gui/src/components/offers2/CreateOfferBuilder.tsx
+++ b/packages/gui/src/components/offers2/CreateOfferBuilder.tsx
@@ -71,13 +71,14 @@ export default function CreateOfferBuilder(props: CreateOfferBuilderProps) {
 
   const handleSubmit = useCallback(
     async (values: OfferBuilderData) => {
-      const { assetsToUnlock, ...localOffer } = await offerBuilderDataToOffer(
-        values,
+      const { assetsToUnlock, ...localOffer } = await offerBuilderDataToOffer({
+        data: values,
         wallets,
-        offers || [],
-        false,
-        true
-      );
+        offers: offers || [],
+        validateOnly: false,
+        considerNftRoyalty: true,
+        allowEmptyOfferColumn: false,
+      });
 
       const assetsRequiredToBeUnlocked = [];
       const assetsBetterToBeUnlocked = [];
--- a/packages/gui/src/util/offerBuilderDataToOffer.ts
+++ b/packages/gui/src/util/offerBuilderDataToOffer.ts
@@ -32,14 +32,24 @@ export type AssetStatusForOffer = {
   relevantOffers: OfferTradeRecordFormatted[];
 };
 
+export type OfferBuilderDataToOfferParams = {
+  data: OfferBuilderData;
+  wallets: Wallet[];
+  offers: OfferTradeRecordFormatted[];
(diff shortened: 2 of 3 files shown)


packages/gui/src/util/offerBuilderDataToOffer.ts
```diff
@@ -58,10 +68,12 @@ export default async function offerBuilderDataToOffer(
...
 
-  const hasOffer = !!offeredXch.length || !!offeredTokens.length || !!offeredNfts.length;
+  if (!allowEmptyOfferColumn) {
+    const hasOffer = !!offeredXch.length || !!offeredTokens.length || !!offeredNfts.length;
 
-  if (!hasOffer) {
-    throw new Error(t`Please specify at least one offered asset`);
+    if (!hasOffer) {
+      throw new Error(t`Please specify at least one offered asset`);
+    }
```
ChiaMineJP, line 76: I think one-sideded offer includes request-only offer.
How about writing like this?
```typescript
  const hasOffer = !!offeredXch.length || !!offeredTokens.length || !!offeredNfts.length;
  const hasRequest = !!requestedXch.length || !!requestedTokens.length || !!requestedNfts.length;
  if(!hasOffer && !hasRequest){
    throw new Error(t`Offer is empty`);
  }

  if (!allowEmptyOfferColumn) {
    if (!hasOffer || !hasRequest) {
      throw new Error(t`Please specify at least one offering/requesting asset`);
    }
  }
```

packages/gui/src/components/offers2/CreateOfferBuilder.tsx
```diff
@@ -71,13 +71,14 @@ export default function CreateOfferBuilder(props: CreateOfferBuilderProps) {
...
+        data: values,
         wallets,
-        offers || [],
-        false,
-        true
-      );
+        offers: offers || [],
+        validateOnly: false,
+        considerNftRoyalty: true,
+        allowEmptyOfferColumn: false,
```
ChiaMineJP, line 80: Is initiating one-sided offer not allowed?

paninaro: Initiating a one-sided offer is allowed when creating a new offer, but that one-sided offer *MUST* be offering some asset. We don't support creating offers that only request assets.

Maybe that will be a feature that we support at some point, but currently the intent is to only support creating an offer if some assets are being offered.

packages/gui/src/util/offerBuilderDataToOffer.ts
paninaro: I checked with E-ming, and the intent is to only support one-sided offers where assets are being offered. It isn't currently planned to support a request-only offer.

ChiaMineJP: In that case, the code looks good to me

packages/gui/src/components/offers2/CreateOfferBuilder.tsx
ChiaMineJP: Ok, thanks for the detail

ChiaMineJP (approved):

PR #1662, opened 2023-04-04
update decode-uri-component transitive dep to 0.2.2

paninaro:
Addresses https://github.com/Chia-Network/chia-blockchain-gui/security/dependabot/48


wjblanke (commented):
aok

PR #1663, opened 2023-04-04
invalidate some RTK caches when logging into a key

paninaro:
The current wallet address wasn't being updated when logging into a newly created key. This change will invalidate several of the RTK caches when the login mutation is invoked, causing the current wallet address to refresh properly, along with wallets/balances/notifications


seeden (approved):
fine as temp solution,
but we need to figure our why api cache reset is not working properly after login

seeden (approved):

PR #1674, opened 2023-04-15
Fixed broken code

ChiaMineJP:
## Fixed an issue where bladebit plotter doesn't work
Simply you couldn't create plots with bladebit from GUI.

## Fixed an issue where fullnode status kept loading even after fully synced
image
The root cause is that `get_blockchain_state` FullNode RPC API is invoked before a connection from `full_node` is registered to daemon. You need to wait for API timeout (10 minutes) to get the loader disappeared.

### Memo
`is_running` API can return true if process of the target service exists but ws connection is not established.
We need to check the result of `ping` to know whether service is listening to the web socket connection.


packages/api/src/Client.ts
```diff
@@ -65,7 +67,7 @@ export default class Client extends EventEmitter {
     super();
 
     this.options = {
-      timeout: 60 * 1000 * 10, // 10 minutes
+      timeout: 30 * 1000, // 30 seconds
```
paninaro, line 70: While I don't like the 10 minute timeout, I think it would cause issues lowering it to 30 seconds. Some RPC calls have historically taken > 30 seconds to respond.

ChiaMineJP: Thanks for the review! I updated the timeout to 90 seconds.

packages/api/src/services/Daemon.ts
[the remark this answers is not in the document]
ChiaMineJP: automatic camelCase to snake_case conversion is not turned on in `start_plotting` command since `disableFormatting` in `this.command(...)` is set to `true` for this command.
You can try bladebit disk plot with the `main` branch. You will encounter the error.
The root cause of the error is `conversionDict` is missing `plotType` key.

I first thought that set `disableFormatting` to `undefined` or `false` will solve this issue but in this case conversion takes place 2 times and it is wasting of cpu. (since only one conversion suffices by modifying `conversionDict`.)

packages/api/src/Client.ts
```diff
@@ -65,7 +67,7 @@ export default class Client extends EventEmitter {
     super();
 
     this.options = {
-      timeout: 60 * 1000 * 10, // 10 minutes
+      timeout: 90 * 1000, // 90 seconds
```
paninaro, line 70: Zlatko suggested that we keep this at 10 minutes, and if necessary, we can look at overriding the timeout on a per-command basis.

I'll revert this change here in the interests of getting a beta build out with this fix.

paninaro, line 70:
```suggestion
      timeout: 60 * 1000 * 10, // 10 minutes
```

paninaro (approved):

PR #1687, opened 2023-04-21
Added address book prototype v2

joytechu:
This is a proposed version of an address book for the chia client that will enable contacts to be added, allow for users to have copy/paste addresses when sending transactions. It also shows a friendly name against a transaction if the transaction comes from a contact in your address book.

This PR is a PoC v2
- Creates contacts and stores them in local storage.
- Allows for Dids, Nft & domains to be associated with a contact.
- Added side menu item for an address book.

Screenshot 2023-04-21 at 14 52 48

Outstanding
-  Editing a contact
- Nft preview was causing failures when nftid was not a valid id, so it's commented out temporarily.
- Showing name in transaction history 

(will be re-added shortly).

CC:
@paninaro


paninaro:
Thanks for the continued effort on this @joytechu! I'm going to merge this into a branch so that we can collaborate on this.

PR #1689, opened 2023-04-21
fix broken calls to getOfferSummary

paninaro:
The broken calls were preventing offers from being viewable.


seeden (approved):

paninaro:
@zsolt-dev for visibility. I'll check the other mutations but I suspect there may be other similar issues.

PR #1700, opened 2023-04-24
Verifiable credentials hooks

rifeljm:


packages/api-react/src/services/index.ts
```diff
@@ -184,6 +184,14 @@ export const {
 
   // verify
   useVerifySignatureMutation,
+
+  // VC
```
paninaro, line 188:
```suggestion
  // VC (Verifiable Credentials)
```

packages/api-react/src/services/wallet.ts
```diff
@@ -1396,6 +1396,31 @@ export const walletApi = apiWithTag.injectEndpoints({
     }),
 
     verifySignature: mutation(build, WalletService, 'verifySignature'),
+
+    getVC: build.query<string, { vcId: string }>({
```
paninaro, line 1400: Can you update this to use the newer `query(build, VC, 'getVC')` syntax? Take a look at `getCATName` as an example.

packages/api-react/src/services/wallet.ts
```diff
@@ -1396,6 +1396,31 @@ export const walletApi = apiWithTag.injectEndpoints({
...
+      query: ({ vcId }) => ({
+        command: 'getVC',
+        service: VC,
+        args: vcId,
+      }),
+      transformResponse: (response) => response.vcRecord,
+    }),
+
+    getVCList: build.query<string, { start: number; count: number }>({
```
paninaro, line 1409: Same here. We should use the `query(build, VC, 'getVCList')` syntax.

paninaro:
@zsolt-dev can you take a look and make any suggestions regarding how to include type information for the RPC requests/responses?

paninaro (approved):

PR #1730, opened 2023-04-30
nft improvements 2

seeden:
fixed cache manager
fixed NFT search input
fixed memory leak
fixed downloading of file
fixed aborted requests
created cache Provider
fixed NFT filters and Provider


packages/gui/src/components/nfts/provider/hooks/useMetadataData.ts
```diff
@@ -0,0 +1,186 @@
...
+
+        // reload metadata
+        getMetadata(id);
+      }
+    },
+    [getMetadata /* immutable */, metadatasOnDemand /* immutable */]
+  );
+
+  // immutable function
+  const subscribeToMetadataChanges = useCallback(
```
paninaro, line 163: Seems like this can only be (properly) used within a `useEffect` since it returns a cleanup function. Looks like the only caller is in useNFTMetadata currently, which calls it from a useEffect hook. I'd be a bit concerned about using it elsewhere inappropriately.

packages/gui/src/components/nfts/provider/hooks/useMetadataData.ts
```diff
@@ -0,0 +1,186 @@
...
+    (nftId: string, metadataOnDemand: MetadataOnDemand) => {
+      log(`Setting metadata on demand for ${nftId}`);
+
+      metadatasOnDemand.set(nftId, metadataOnDemand);
+
+      events.emit(getChangedEventName(nftId), {
+        metadata: metadataOnDemand.metadata,
+        error: metadataOnDemand.error,
+        isLoading: !!metadataOnDemand.promise,
+      });
```
paninaro, line 47: Is the idea here that instead of using a state-changing `setMetadatasOnDemand`, we modify `metadatasOnDemand` directly to avoid re-rendering and instead rely on sending an event to triggering any re-rendering?

packages/gui/src/components/nfts/provider/hooks/useNFTData.ts
```diff
@@ -0,0 +1,295 @@
...
+    return fetchData(nftWallets, abortControllerRef.current.signal);
+  }, [fetchData, nftWallets]);
+
+  useEffect(() => {
+    // abort previous fetch
+    if (fingerprint) {
+      let changed = false;
+      let initialLoadFinished = false;
+
+      refetchData().finally(() => {
```
paninaro, line 206: I haven't tested this, but if `refetchData()` calls `abort()`, does it throw an exception? If so, should that be caught and handled here?

paninaro: Same question would apply to the other callers of `refreshData()` below

packages/gui/src/components/nfts/provider/hooks/useNFTData.ts
```diff
@@ -0,0 +1,295 @@
...
+export default function useNFTData(props: UseNFTDataProps) {
+  const { pageSize = 24, concurrency = 10 } = props;
+
+  const abortControllerRef = useRef<AbortController>(new AbortController());
+  const [nfts] = useState(() => new Map<string, NFTInfo>());
+  const add /* immutable until concurrency change */ = useMemo(() => limit(concurrency), [concurrency]);
+
+  const events /* immutable */ = useMemo(() => {
+    const eventEmitter = new EventEmitter();
+    eventEmitter.setMaxListeners(Infinity);
```
paninaro, line 37: ♾️ seems like a lot :). Will setting this make it more difficult to troubleshoot legitimate issues?

packages/gui/src/components/signVerify/SigningEntityDID.tsx
```diff
@@ -63,40 +63,33 @@ export default function SigningEntityDID(props: SigningEntityDIDProps) {
...
-                    </InputAdornment>
-                  ),
-                }}
-                fullWidth
-              />
-            </Flex>
-          </Box>
-        </Box>
-      </Grid>
+      <DIDProfileDropdown
```
paninaro, line 66: Thanks for fixing the layout issues here

packages/gui/src/hooks/useFetchAndProcessMetadata.ts
```diff
@@ -0,0 +1,37 @@
...
+      if (hash && !compareChecksums(checksum, hash)) {
+        throw new Error('Checksum mismatch');
+      }
+
+      const headers = await getHeaders(uri);
+      const content = await getContent(uri);
+
+      const metadataString = parseFileContent(content, headers);
+
+      return JSON.parse(metadataString) as Metadata;
```
paninaro, line 31: Any concerns about the metadata not being valid JSON? If this throws, is the caller expected to handle the exception?

packages/gui/src/hooks/useNFT.ts
```diff
@@ -1,34 +1,29 @@
...
-    events.on('nftChanged', handleChange);
-    setNFTState(getNft(nftId));
-
-    return () => {
-      events.off('nftChanged', handleChange);
-    };
-  }, [events, getNft, nftId]);
+  useEffect(
+    () =>
+      subscribeToNFTChanges(id, (newNFTState) => {
```
paninaro, line 22: Related to my earlier comment, should we at least include a comment here that `subscribeToNFTChanges` returns a cleanup function?

packages/gui/src/hooks/useNFTCoinEvents.ts
```diff
@@ -0,0 +1,43 @@
...
+    (callback: (event?: Event) => void) => {
+      events.on('change', callback);
+      return () => {
+        events.off('change', callback);
+      };
+    },
+    [events]
+  );
+
+  return subscribe;
```
paninaro, line 42: Another case that concerns me about improper usage.

packages/gui/src/hooks/useNFTMetadata.ts
```diff
@@ -1,34 +1,29 @@
...
 
-    events.on('metadataChanged', handleChange);
-    setMetadataState(getMetadata(nftId));
-
-    return () => {
-      events.off('metadataChanged', handleChange);
-    };
-  }, [events, getMetadata, nftId]);
+  useEffect(
+    () =>
```
paninaro, line 21: Should we add a comment about the returned cleanup function?

packages/gui/src/components/nfts/provider/hooks/useNFTData.ts
paninaro: I see Infinity used elsewhere. My primary concern is that if there's a problem with something registering too many listeners, it may be difficult to track down the issue.

packages/gui/src/hooks/useFetchAndProcessMetadata.ts
seeden: no, we are handling errors in functions which are using this function

packages/gui/src/components/nfts/provider/hooks/useNFTData.ts
seeden: refetchData never emit any error because it is using setError

packages/gui/src/components/nfts/provider/hooks/useMetadataData.ts
seeden: useMetadataData is an internal hook used only in Provider. 
therefore it is not inside src/hooks directory

packages/gui/src/components/nfts/provider/hooks/useMetadataData.ts
seeden: yes, it is performance improvement. only components which are using  hook useNFTMetadata will be reloaded

packages/gui/src/hooks/useNFTCoinEvents.ts
seeden: I do not understand what is wrong with this one ?

packages/gui/src/components/nfts/provider/hooks/useNFTData.ts
seeden: We have on one screen 50 nft card, each card can have 5+ useNFTMetadata hooks => 5x50 = 250 events.
Notifications screen are using useNFTMEtadata as well => 4 * 10 = ~ 50
NFTAutocomplete can use more.
We cam add limit but I do not know what is a good number. 
Maybe 1000 ?

paninaro: I'd be fine with a large number like 10_000. I don't think we need to hold this PR up for this change though.

paninaro (approved):

PR #1745, opened 2023-05-04
add color constant

lipalong:
Add color constants for updated color theming. Update light and dark mode wallet color themes to use color constants.


paninaro (approved):
lgtm

PR #1759, opened 2023-05-10
generate test coverage report merged from all packages

paninaro:
We already run jest across all packages with the --coverage option. This change will save the coverage output for each package and then merge it into a single report. We'll explore publishing the coverage report separately.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,9 @@
     "locale": "lerna run locale",
     "audit": "lerna exec -- npm audit",
     "audit:fix": "lerna exec -- npm audit fix",
-    "test": "lerna run test --stream -- -- --passWithNoTests --coverage",
+    "test": "lerna run test --stream -- -- --passWithNoTests --coverage --coverageDirectory=.coverage && npm run coverage:merge && npm run coverage:report",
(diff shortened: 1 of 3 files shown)


wallentx (approved):

package.json
```diff
@@ -29,7 +29,9 @@
     "locale": "lerna run locale",
     "audit": "lerna exec -- npm audit",
     "audit:fix": "lerna exec -- npm audit fix",
-    "test": "lerna run test --stream -- -- --passWithNoTests --coverage",
+    "test": "lerna run test --stream -- -- --passWithNoTests --coverage --coverageDirectory=.coverage && npm run coverage:merge && npm run coverage:report",
+    "coverage:merge": "npx istanbul-merge --out .nyc_output/coverage.json packages/api/.coverage/coverage-final.json packages/api-react/.coverage/coverage-final.json packages/core/.coverage/coverage-final.json packages/gui/.coverage/coverage-final.json  packages/icons/.coverage/coverage-final.json packages/wallets/.coverage/coverage-final.json",
+    "coverage:report": "npx nyc report --reporter=html --reporter=text --reporter=lcov",
```
ChiaMineJP, line 34: `--reporter=lcov` is `--reporter=lcov-only` + `--reporter=html`.
So `--reporter=html` isn't necessary unless you want to have the report files under `.coverage/` directory as well as `.coverage/lcov-report/`. (The same html files would be genereated to both directory)

ChiaMineJP (dismissed):

package.json
paninaro: Thanks, I'll make that change. This was a GitHub Copilot suggestion :)

wallentx (approved):

PR #1761, opened 2023-05-10
Fix refetching plots after new plot is finished plotting

### LINKED ISSUE #1728: [Bug] New Plot not displayed after it finishes plotting
### What happened?

1. create a new plot
2. wait for it to finish
3. Plot list is not refreshed (the new one is missing)

It seems like the plot list is refreshed (visible flicker/refresh after plotting finished) but its not yet returned by the API. 

### Version

branch release/1.8.0

### What platform are you using?

macOS

### What ui mode are you using?

GUI

### Relevant log output or stacktrace

_No response_

zsolt-dev:
fixes: https://github.com/Chia-Network/chia-blockchain-gui/issues/1728

What other solutions were considered:

1. Have the logic of clearing cache directly in the api-react
-> spend hours on it, could not make it work. Queries should not clear caches, mutations should...

2. Have the logic of clearing cache in useGetThrottlePlotQueueQuery
-> might be a better place, but this is obvious for the developer that some cache cleaning is happening. I can move it to useGetThrottlePlotQueueQuery if someone thinks its a better place.

Thank you

### DIFF
--- a/packages/api-react/src/hooks/useGetThrottlePlotQueueQuery.ts
+++ b/packages/api-react/src/hooks/useGetThrottlePlotQueueQuery.ts
@@ -1,5 +1,7 @@
 import { PlotQueueItem } from '@chia-network/api';
+import { useEffect, useMemo } from 'react';
 
+import { useRefreshPlotsMutation } from '../services/harvester';
 import { useGetPlotQueueQuery } from '../services/plotter';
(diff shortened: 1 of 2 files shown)


seeden:
@zsolt-dev can we move it to useGetThrottlePlotQueueQuery?

PR #1785, opened 2023-05-23
Rifeljm/#1699 verifiable credentials

rifeljm:


packages/api/src/wallets/VC.ts
```diff
@@ -101,22 +101,25 @@ export default class VCWallet extends Wallet {
...
       proofs: any;
-    }>('add_vc_proofs', { proofs });
+    }>('vc_add_proofs', { proofs });
   }
 
   async getProofsForRoot(root: string) {
-    return this.command<{ proofs: any }>('get_proofs_for_root', { root });
+    if (!root) {
+      return { proofs: {} };
+    }
```
paninaro, line 116: We should let the backend respond even if the request is malformed. The backend could change how it handles this scenario.

paninaro (approved):

Issue #1786, opened 2023-05-23
[BUG] Cannot read properties of undefined (reading 'toString')


dkackman:
### What happened?

Attempting to view the detail page or select Move to Profile of this NFT:
https://www.spacescan.io/xch/nft/nft1e0wmps6u02fuw872cjtynh8utry4qgzned2rlszmcfux4ett4k3s4uz63z

It is an audio file and perhaps unrelated but due to a bug in the code I used to create it, the file name ends in `..mpga` (not two dots).

### Version

1.8.1

### What platform are you using?

Windows

### What ui mode are you using?

GUI

### Relevant log output or stacktrace

```shell
Cannot read properties of undefined (reading 'toString')

URL
#/dashboard/nfts

Stacktrace
at webpack://@chia-network/gui/src/components/nfts/NFTProperties.tsx:44:25 NFTProperty
at webpack://@chia-network/gui/node_modules/react-dom/cjs/react-dom.production.min.js:157:136 Ch
at webpack://@chia-network/gui/node_modules/react-dom/cjs/react-dom.production.min.js:267:459 b?b.sibling:null}I
at webpack://@chia-network/gui/node_modules/react-dom/cjs/react-dom.production.min.js:250:346 Pj
at webpack://@chia-network/gui/node_modules/react-dom/cjs/react-dom.production.min.js:250:277 Pj
at webpack://@chia-network/gui/node_modules/react-dom/cjs/react-dom.production.min.js:250:137 Pj
at webpack://@chia-network/gui/node_modules/react-dom/cjs/react-dom.production.min.js:243:162 Lj
at webpack://@chia-network/gui/node_modules/react-dom/cjs/react-dom.production.min.js:123:114 jg
at webpack://@chia-network/gui/node_modules/scheduler/cjs/scheduler.production.min.js:18:342 unstable_next
at webpack://@chia-network/gui/node_modules/react-dom/cjs/react-dom.production.min.js:122:324 eg
```

PR #1789, opened 2023-05-24
fix loading of some audio NFTs

paninaro:
On behalf of @dkackman:
For an audio NFT (or at least the one indicated in the #1786)

rawValue on line 44 of \chia-blockchain-gui\packages\gui\src\components\nfts\NFTProperties.tsx can be undefined. Added a simple check alongside where typeof object is checked and return null. Haven't seen downstream consequences and the nft detail page now opens without error.

value on that line could also be set to an empty string or other default but I don't know enough about the code to know what is the best approach so deiced an early bail out, since it already is in there, was safest.


seeden (approved):

dkackman:
thanks! this encouraged me to finally set up commit signing by default

PR #1813, opened 2023-06-01
replace isURL usage with isValidURL

paninaro:
Some NFTs have content URLs containing unencoded spaces. While most of the code handles this without any trouble, the isURL function from the validator library will return false, causing these NFTs to fail to load.

This change introduces an `isValidURL` function that will encode URLs if necessary and then return the result of `isURL(encodedURL)`.

Tested with NFTs that have encoded and unencoded spaces.

### DIFF
--- a/packages/gui/src/electron/CacheManager.ts
+++ b/packages/gui/src/electron/CacheManager.ts
@@ -305,8 +305,15 @@ export default class CacheManager extends EventEmitter {
 
     const process = async (): Promise<CacheInfo> => {
       try {
-        if (!isURL(url)) {
-          throw new Error(`Invalid URL: ${url}`);
+        // From isValidURL.ts
+        // isURL returns false for URLs with unencoded spaces. We can't use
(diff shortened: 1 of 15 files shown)


packages/gui/src/electron/CacheManager.ts
```diff
@@ -305,8 +305,15 @@ export default class CacheManager extends EventEmitter {
...
     const process = async (): Promise<CacheInfo> => {
       try {
-        if (!isURL(url)) {
-          throw new Error(`Invalid URL: ${url}`);
+        // From isValidURL.ts
+        // isURL returns false for URLs with unencoded spaces. We can't use
+        // encodeURI if the URL is already encoded, so we attempt to decode
+        // the URL first and then encode it if it wasn't already encoded.
+
+        const normalizedURL = decodeURI(url) === url ? encodeURI(url) : url;
```
seeden, line 313: we can use here isValidURL as well

seeden (changes requested):

seeden (approved):

PR #1815, opened 2023-06-01
fixed switching between keys

seeden:
fixed push notifications during switching between keys

### DIFF
--- a/packages/core/src/components/Auth/AuthProvider.tsx
+++ b/packages/core/src/components/Auth/AuthProvider.tsx
@@ -41,7 +41,7 @@ export default function AuthProvider(props: AuthProviderProps) {
     [clearCache]
   );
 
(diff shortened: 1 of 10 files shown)


packages/core/src/components/Auth/AuthProvider.tsx
```diff
@@ -0,0 +1,96 @@
...
+          setFingerprint(newFingerprint);
+        } finally {
+          setIsLoading(false);
+        }
+      }
+    },
+    [clearCache]
+  );
+
+  // automatically log in if we have a fingerprint already and not logIn is not in progress
```
paninaro, line 44:
```suggestion
  // automatically log in if we have a fingerprint already and logIn is not in progress
```

paninaro (approved):

PR #1839, opened 2023-06-14
Zsolt/clawback enhancements

zsolt-dev:
Please read the individual commit messages for more info


paninaro:
We should apply "Change selectedTab to summary after sending a TX" when sending a CAT as well. I'll create an issue.

paninaro (approved):

PR #1849, opened 2023-06-15
Rifeljm/vc is valid lazy get proofs

rifeljm:


packages/gui/src/components/vcs/VCList.tsx
```diff
@@ -83,6 +84,9 @@ export default function VCList() {
 
   const [proofs, setProofs] = React.useState<any>({});
 
+  /* We only need to subscribe to event and the list will be updated automatically on added VC */
+  useVCCoinAdded(() => {});
```
paninaro, line 88: Are we sure this isn't creating a leak? Looks to me like this needs to be wrapped in a useEffect that returns a cleanup/unsubscribe function.

@seeden can you provide some guidance here?

paninaro: We can address this in the next release.

paninaro (approved):

PR #1865, opened 2023-06-22
updated mui and lingui

seeden:
removed unused package-lock files
fixed error :(0 , muibase__WEBPACK_IMPORTED_MODULE_0__.generateUtilityClasses) is not a function upgraded lingui => we can use msq`` instead defineMessage() and try to add missing language


paninaro (approved):
Tested, working.

Issue #1862, opened 2023-06-21
New wallet doesn't load automatically [Bug] 


danieljperry:
### What happened?

After creating a new wallet, the GUI displays the previous wallet instead of the new one. See video attachment:

https://github.com/Chia-Network/chia-blockchain-gui/assets/69756004/7a215ea1-e620-4dbd-8b30-c04233554435

To reproduce:
1. say you are viewing Alice's wallet, exit that wallet to the primary Chia Wallet Keys dialog
2. Add a new wallet
3. At this point, I would expect the new wallet to be displayed. Instead, Alice's wallet is displayed
4. Exit Alice's wallet; this verifies that the new wallet actually does exist, but wasn't loaded by default

### Version

1.8.2rc4 + PR 15496

### What platform are you using?

Windows

### What ui mode are you using?

GUI

paninaro:
Thanks for the report @danieljperry. We have a separate item tracking this, but I'll keep this around for visibility.

seeden:
Fixed via https://github.com/Chia-Network/chia-blockchain-gui/pull/1868/files

PR #1868, opened 2023-06-23
used new useAuth hook for logIn

seeden:
used new useAuth hook for logIn and fixed https://github.com/Chia-Network/chia-blockchain-gui/issues/1862


paninaro (approved):
Tested wallet creation via "Create New" and "Import Existing". Works!

PR #1869, opened 2023-06-23
updated wc

seeden:
used latest version of wallet connect


paninaro:
I'm not able to successfully create a session with a dapp. The pairing appears to work, and the dapp appears to connect, but the GUI isn't showing any active sessions. Any request sent by the dapp results in an invalid request error.

PR #1890, opened 2023-07-06
bugfix: crash importing local VC

rifeljm:


packages/gui/src/components/vcs/VCCard.tsx
```diff
@@ -44,7 +43,7 @@ export default function VCCard(props: { vcRecord: any; isDetail?: boolean; proof
   const [revokeVC] = useRevokeVCMutation();
   const theme: any = useTheme();
   const openDialog = useOpenDialog();
-  const [vcTitlesObject] = usePrefs<any>('verifiable-credentials-titles', {});
+  const [vcTitlesObject] = useLocalStorage<any>('verifiable-credentials-titles', {});
```
paninaro, line 46: Why are we switching from prefs to local storage? Won't we lose existing data from prefs if we make this change?

packages/gui/src/components/vcs/VCCard.tsx
```diff
@@ -55,6 +54,7 @@ export default function VCCard(props: { vcRecord: any; isDetail?: boolean; proof
   const [pendingRevoke, setPendingRevoke] = useLocalStorage<any>('verifiable-credentials-pending-revoke', {});
   const { data: fingerprint } = useGetLoggedInFingerprintQuery();
   const [getTransactionAsync] = useGetTransactionAsyncMutation();
+  const isVCLocal = React.useMemo(() => !!vcRecord.isLocal, [vcRecord.isLocal]);
```
paninaro, line 57: Why is this using useMemo? Also, shouldn't this be checking if vcRecord is undefined? We check `vcRecord?.<property>` elsewhere.

packages/gui/src/components/vcs/VCCard.tsx
```diff
@@ -84,10 +84,14 @@ export default function VCCard(props: { vcRecord: any; isDetail?: boolean; proof
   }
 
   function renderProperties() {
-    const didString = vcRecord?.vc?.proofProvider
-      ? didToDIDId(vcRecord?.vc?.proofProvider)
-      : vcRecord.credentialSubject?.id || '/';
-
+    let didString: string = vcRecord.credentialSubject?.id || '/';
```
paninaro, line 87: The use of `vcRecord.` and `vcRecord?.` is inconsistent in this function.

packages/gui/src/components/vcs/VCCard.tsx
```diff
@@ -84,10 +84,14 @@ export default function VCCard(props: { vcRecord: any; isDetail?: boolean; proof
...
+    try {
+      if (vcRecord?.vc?.proofProvider) {
+        didString = didToDIDId(vcRecord?.vc?.proofProvider);
```
paninaro, line 90: nit: assign `vcRecord?.vc?.proofProvider` to a local variable instead of copying & pasting

packages/gui/src/components/vcs/VCCard.tsx
paninaro: Maybe `isLocal` should be a VCCard prop instead of attaching the property to vcRecord.

paninaro (changes requested):
Please take a look at my comments

packages/gui/src/components/vcs/VCCard.tsx
rifeljm: A huge mistake on my end. I thought if we're storing VCs in localStorage, then titles shouldn't be in prefs.
However I completely forgot about onchain VCs :/

packages/gui/src/components/vcs/VCCard.tsx
rifeljm: Fixed.

packages/gui/src/components/vcs/VCCard.tsx
rifeljm: Fixed.

paninaro (approved):

PR #1899, opened 2023-07-14
Refactoring playwright scripts to point to designated wallet

jfawChia:
This merge happen without successfully completing all ESLINT suggestions/requirements due to some that would break how playwright interacts with the Chia app. 

I REMOVE THE FOLLOWING: 
1. Console statements
2. Whitespaces
3. Unused declarations 

THE FOLLOWING WILL MESS UP WORKING CODE: 
1. 'playwright' should be listed in the project's dependencies, not optionalDependencies 
2.  Some console.log are apart of the test expectation output and not for debugging
3. Changing variable in send_amount_greater.spec from let to const through errors upon execution. Need to leave as let declaration.


.eslintrc.json
```diff
@@ -102,7 +102,7 @@
           // You can change the file name extension to .dev.ts to avoid the: "XX should be listed in the project's dependencies, not devDependencies."
           "**/*.dev.{js,jsx,ts,tsx}"
         ],
-        "optionalDependencies": false
+        "optionalDependencies": true
```
paninaro, line 105: @zsolt-dev Are you ok with this change?

paninaro: Should be fine. Resolving...

paninaro (approved):

PR #1901, opened 2023-07-16
fix walletConnect mintNFT error handling

BrandtH22:
update to ensure errors are properly passed through the mintNFT walletConnect command


seeden:
@BrandtH22 I do not understand how this will fix your issue. Please provide detailed explanation. 
All errors are send back to Dapp with https://github.com/Chia-Network/chia-blockchain-gui/blob/878447b25d4a9163767e712938aaf7d69769fb32/packages/gui/src/util/walletConnect.ts#L245

PR #1916, opened 2023-07-20
add address book integration to send CAT tx

lipalong:
- Add address book integration to the address field on the send CATs transaction page
- Clean up


paninaro (approved):
Looks good, thanks!

PR #1931, opened 2023-07-24
Color fixes in gui

lipalong:
Fixed color issues in the GUI in the wallet tx list (when using clawback) and on the verifiable credentials pages


paninaro (approved):
Looks good. Thanks for the fix!

PR #1934, opened 2023-07-24
Add warning text when sending clawback TX

zsolt-dev:
https://github.com/Chia-Network/chia-blockchain-gui/assets/18726534/9f104d34-9adb-4c2a-bab7-d91a4df96162

@sethjenks If you have some better design, please send it in. Thank you

### DIFF
--- a/packages/wallets/src/components/WalletSend.tsx
+++ b/packages/wallets/src/components/WalletSend.tsx
@@ -23,6 +23,7 @@ import {
 import { ConnectCheckmark } from '@chia-network/icons';
 import { Trans, t } from '@lingui/macro';
 import ExpandMoreIcon from '@mui/icons-material/ExpandMore';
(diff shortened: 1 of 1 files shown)


BrandtH22:
I really like this addition!

Out of curiosity, since a number of users have now setup auto-clawback in the settings would it make sense to have the warning as a popup that the user must confirm after clicking the send button and before the clawback transaction is actually sent? (There could a check box on it to hide the popup in the future and accept the potential issues that could arise)

paninaro:
@BrandtH22 I'm not opposed to displaying a popup, but for now I'm satisfied with this solution. Will follow up with Seth.

paninaro:
Cherry-picked these changes into a separate PR targeting `release/2.0.0`: https://github.com/Chia-Network/chia-blockchain-gui/pull/1937

PR #1954, opened 2023-07-27
Fixed add plot page

ChiaMineJP:
- Fixed an issue where farmer public key is emptified after changing plotter
- Fixed an issue where manual key setup didn't work even if the checkbox is on
- Shows error when farmer/pool public key is invalid when they are manually entered
- Add current logged-in farmer/pool pk hints to manual key inputs


paninaro (approved):
Looks good to me. Thanks for fixing this.

PR #1959, opened 2023-07-31
Hide GPU setting when it is not available

ChiaMineJP:
Hide GPU settings not only Mac but also computers on which cudaplot is not available.


paninaro (approved):
lgtm

PR #1979, opened 2023-08-05
Plot filter value now is the function of height

ChiaMineJP:
Before: plot filter is a constant (1/512)
After: plot filter changes by height.

image


danieljperry (approved):
lgtm

PR #1987, opened 2023-08-08
Added compressed plot size texts

ChiaMineJP:
Source: https://docs.chia.net/k-sizes/

image

Also this PR hides compress option for bladebit_disk.

### DIFF
--- a/packages/gui/src/constants/plotSizes.ts
+++ b/packages/gui/src/constants/plotSizes.ts
@@ -7,6 +7,18 @@ type PlotSize = {
   defaultRam: number;
 };
 
+export const compressedSizes: Record<number, Record<number, string>> = {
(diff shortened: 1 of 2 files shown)


packages/gui/src/constants/plotSizes.ts
```diff
@@ -43,9 +55,23 @@ export const plottingInfo: Record<PlotterName, PlotSize[]> = {
...
+  return plottingInfo[plotterName].map((item) => {
+    const kSize = item.value;
+    if (
+      typeof compressionLevel !== 'number' ||
+      !compressedSizes[compressionLevel] ||
+      !compressedSizes[compressionLevel][kSize]
+    ) {
+      return {
+        value: kSize,
+        label: `k=${kSize} (Effective plot size: ${item.effectivePlotSize}, Temporary space: ${item.workspace})`,
```
paninaro, line 68: Is this label being localized somewhere else? Same question for the label on line 74

PR #1997, opened 2023-08-12
Fixed an issue where plotNFT was not created after changing plotter

ChiaMineJP:
There is currently a bug where creating plotNFT is silently canceled after changing plotter.
The root cause is when changing plotter, form values are reset to plotter's default values.
Since the default value of create-nft request is `false`, even if filling plotNFT form, changing plotter reset the flag for creating plotNFT as `false`.

image

image


ChiaMineJP:
This PR also closes this issue https://github.com/Chia-Network/chia-blockchain/issues/15980

paninaro (approved):

PR #2037, opened 2023-08-25
Fix address entry when entered directly instead of using the selection

paninaro:
Entering an address via typing/pasting was broken by the removal of the `autoSelect` prop. That prop was removed to prevent automatic selection on blur if the user didn't actually make a selection. This fix updates the TextField's props to include an onChange handler that is invoked when typing/pasting.


lipalong (approved):
lgtm

PR #2047, opened 2023-08-28
CR-CAT offers functionality

zsolt-dev:
Provided a video with steps in the chat.


packages/core/src/constants/CrCatsWellKnownProviders.ts
```diff
@@ -1,5 +1,5 @@
 const CrCatsWellKnownProviders = {
-  dafa583812b792d00f3a05db054d99ea4b82d7203b525c298c96069fb469d4aa: {
+  a8f06692e013166feffb0fa10db1f2d594350a84e0dcbb8ad8c156f00bb86c8f: {
```
paninaro, line 2: Can you add a comment that this is for testing only?

packages/api-react/src/services/wallet.ts
```diff
@@ -79,7 +79,7 @@ export const walletApi = apiWithTag.injectEndpoints({
               wallets.map(async (wallet: Wallet) => {
                 const { type } = wallet;
                 const meta: any = {};
-                if (type === WalletType.CAT) {
+                if ([WalletType.CAT, WalletType.CRCAT].includes(type)) {
```
paninaro, line 82: Since this appears to be repeated in quite a few places, do you think it would make sense to write a small function to encapsulate this logic?

paninaro (changes requested):
Can you remove the localized messages.po files from the PR? Or did you intentionally include them?

paninaro (approved):

PR #2050, opened 2023-08-29
removed incompatible react suspender

seeden:
fixed settings profiles
fixed list of blocks
removed incompatible react suspender with react 18


paninaro (approved):
lgtm. Tested the profiles screen.

PR #2051, opened 2023-08-29
getPublicKey

rifeljm:


packages/gui/src/constants/WalletConnectCommands.tsx
```diff
@@ -1001,6 +1001,20 @@ const walletConnectCommands: WalletConnectCommand[] = [
       },
     ],
   },
+  {
+    command: 'getPublicKey',
+    label: <Trans>Get public key</Trans>,
+    description: <Trans>Requests a master public key from your wallet</Trans>,
+    service: ServiceName.DAEMON,
+    bypassConfirm: true,
```
paninaro, line 1009: This needs to be removed. The documented feature request states that this command should always prompt the user.

packages/gui/src/constants/WalletConnectCommands.tsx
```diff
@@ -1001,6 +1001,20 @@ const walletConnectCommands: WalletConnectCommand[] = [
...
+    params: [
+      {
+        name: WalletConnectCommandParamName.FINGERPRINT,
```
paninaro, line 1012: I hit an issue while testing. I paired with wallet key `3429533387`, and then issued a chia_getPublicKey request with fingerprint `4091547012`. The request failed with the error `"Fingerprint not found"`, even though I do have the wallet key with fingerprint `4091547012`.

The issue is here: https://github.com/Chia-Network/chia-blockchain-gui/blob/main/packages/gui/src/util/walletConnect.ts/#L229-L231
```
    if (!pair.fingerprints.includes(updatedParams.fingerprint)) {
      throw new Error('Fingerprint not found');
    }
```

So we have a parameter name conflict using `fingerprint`, since our WalletConnect code treats that parameter as a special case. Perhaps we should change the parameter name for this to something like `key_fingerprint` and use the `'EXECUTE'` code that Zlatko added to directly call the `useGetPublicKey` RPC hook passing in the `key_fingerprint` value for `fingerprint`.

paninaro (changes requested):
Need to remove bypassConfirm and address the `fingerprint` issue discovered while testing.

packages/gui/src/constants/WalletConnectCommands.tsx
paninaro: Clarified with product folks that this behavior is ok. The Dapp has access to the fingerprint(s) of the paired wallet keys, so passing in the fingerprint of a paired wallet should work properly.

paninaro (approved):

PR #2055, opened 2023-08-31
Rifeljm/#622 walletconnect bypass readonly commands

rifeljm:


packages/gui/src/components/walletConnect/WalletConnectAddConnectionDialog.tsx
```diff
@@ -212,6 +221,23 @@ export default function WalletConnectAddConnectionDialog(props: WalletConnectAdd
...
+        sx={{ cursor: 'pointer' }}
+        alignItems="center"
+        onClick={() => {
+          toggleBypassCheckbox(!bypassCheckbox);
+        }}
+      >
+        <Checkbox checked={bypassCheckbox} disableRipple sx={{ paddingLeft: 0 }} />
+        <Typography variant="body2">
+          <Trans>Bypass confirmation for all read-only commands</Trans>
+        </Typography>
```
paninaro, line 236: Tweak wording and add an informative tooltip on hover.

⚠️ NOTE: `TooltipIcon` needs to be imported from `@chia-network/core` ⚠️

```suggestion
        <Flex flexDirection="row" alignItems="center" gap={1}>
          <Typography variant="body2">
            <Trans>Skip confirmation for all read-only commands</Trans>
          </Typography>
          <TooltipIcon>
            <Trans>
              By default, a prompt will be presented each time a WalletConnect command is invoked. Select this option if
              you would like to skip the prompt for all read-only WalletConnect commands.
              <p />
              Commands that create a transaction or otherwise modify your wallet will still require confirmation.
            </Trans>
          </TooltipIcon>
        </Flex>
      </Flex>
```

packages/gui/src/components/settings/SettingsIntegration.tsx
```diff
@@ -131,6 +135,9 @@ export default function SettingsIntegration() {
     }
   }, [topic, pairs, refreshBypassCommands]);
 
+  const readonlySkipValue =
+    bypassReadonlyCommands && topic && fingerprint ? bypassReadonlyCommands[topic][fingerprint] : false;
```
paninaro, line 139: This line crashes if `bypassReadonlyCommands[topic] === undefined`. Easily repro'd if you create a WC pairing without enabling the new bypass option and then select the newly paired dapp in the settings screen.

paninaro:
```suggestion
  const readonlySkipValue =
    bypassReadonlyCommands && topic && fingerprint && bypassReadonlyCommands[topic]
      ? bypassReadonlyCommands[topic][fingerprint]
      : false;
```

paninaro:
- [ ] Can we move the new "Skip Confirmation For All Readonly Commands" into the same box as "Skip Confirmation for Commands"?
- [ ] Let's be consistent with the "readonly"/"read-only" wording. In the pairing dialog we use "read-only" but here it's "readonly". Let's go with "read-only".
- [ ] If "Always Skip" is selected, we should hide the `commands` that are listed under the "Skip Confirmation for Commands" section.

image

paninaro:
The new `bypass-readonly-commands` localStorage item needs to be cleaned up properly when Disconnect, Restore, or Reset buttons are pressed:

⚠️ NOTE: The Disconnect and Restore buttons should only remove the selected dapp's topic from `bypass-readonly-commands`. The Reset button should completely delete the localStorage item. ⚠️

image

packages/gui/src/components/settings/SettingsIntegration.tsx
```diff
@@ -45,6 +46,9 @@ export default function SettingsIntegration() {
     useWalletConnectPairs();
 
   const pairs = get();
+  const { data: fingerprint } = useGetLoggedInFingerprintQuery();
+
+  const [bypassReadonlyCommands, setBypassReadonlyCommands] = useLocalStorage<any>('bypass-readonly-commands', {});
```
paninaro, line 51: Is there a reason not to use the `useWalletConnectPairs` or `useWalletConnectPreferences` hooks? Creating a new localStorage item is going to make maintenance more complex. At the very least we should make the naming of these localStorage items consistent -- maybe `walletConnectBypassReadonlyCommands`. But I would prefer using the existing hooks if possible.

packages/gui/src/components/settings/SettingsIntegration.tsx
```diff
@@ -334,6 +341,31 @@ export default function SettingsIntegration() {
                 </Box>
               )}
             </Flex>
+            <Box {...borderStyle}>
+              <Typography variant="h6">
+                <Trans>Skip Confirmation For All Readonly Commands</Trans>
```
paninaro, line 346: `Read-only` for consistency

packages/gui/src/components/settings/SettingsIntegration.tsx
```diff
@@ -334,6 +341,31 @@ export default function SettingsIntegration() {
...
+                      value={readonlySkipValue}
+                      onChange={(evt: PointerEvent) => {
+                        setBypassReadonlyCommands((localBypassReadonlyCommands) => ({
+                          ...localBypassReadonlyCommands,
+                          [topic]: { [fingerprint]: evt.target.value },
+                        }));
+                      }}
+                    >
+                      <MenuItem value>
+                        <Trans>Always Skip</Trans>
```
paninaro, line 359:
```suggestion
                        <Trans>Always Allow</Trans>
```

packages/gui/src/components/walletConnect/WalletConnectAddConnectionDialog.tsx
```diff
@@ -118,6 +120,13 @@ export default function WalletConnectAddConnectionDialog(props: WalletConnectAdd
     }
 
     const topic = await pair(uri, selectedFingerprints, mainnet);
+    if (bypassCheckbox) {
+      let skipReadOnlyObject = {};
+      selectedFingerprints.forEach((f: number) => {
+        skipReadOnlyObject = { ...skipReadOnlyObject, [f]: true };
+      });
+      setBypassReadonlyCommands({ ...bypassReadonlyCommands, [topic.toString()]: skipReadOnlyObject });
```
paninaro, line 128: It seems odd to create a nested data structure to hold a boolean. The data structure could be simplified to:
```
{
  <topic_1>: [fingerprint_1, fingerprint_2],
  <topic_2>: [fingerprint_3]
}
```

If the fingerprint is in the list, bypass confirmation is enabled. If it's not, then confirmation is required.

packages/gui/src/hooks/useWalletConnectCommand.tsx
```diff
@@ -172,24 +175,28 @@ export default function useWalletConnectCommand(options: UseWalletConnectCommand
...
-        ),
-      params: definitionParams,
-      values,
-      fingerprint,
-      isDifferentFingerprint,
-      command,
-      bypassConfirm,
-      onChange: handleChangeParam,
-    });
+    const isReadOnlyEnabled = bypassReadonlyCommands?.[pairedTopic]?.[fingerprint];
```
paninaro, line 178: Can we rename this to something like `skipConfirmation`?

paninaro (changes requested):
Some edge cases and UX refinement to be ironed-out.

paninaro:
I'm seeing issues with the latest changes. Setting the checkbox to bypass readonly commands during pairing doesn't seem to take effect -- the walletConnectPreferences localStorage item isn't updated.

PR #2065, opened 2023-09-06
VC detail: show Launcher ID and Coin ID

rifeljm:


packages/gui/src/components/vcs/VCCard.tsx
```diff
@@ -38,6 +38,7 @@ function RenderProperty(props: RenderPropertyProps) {
 
 export default function VCCard(props: { vcRecord: any; isDetail?: boolean; proofs?: any; isLocal: boolean }) {
   const { vcRecord, isDetail, proofs, isLocal } = props;
+  const [vCCoinIds] = useLocalStorage<any>('vc-coin-ids', {});
```
paninaro, line 41: I'd like to avoid using local storage for this purpose. Since we're using `useNavigate` to navigate between VCCard and VCDetail, we can pass additional state to include the coinId. You can look at OfferImport.tsx which has the following example:

```
      const navigationPath = offerContainsAssetOfType(offerSummary, 'singleton')
        ? '/dashboard/offers/view-nft'
        : '/dashboard/offers/view';

      navigate(navigationPath, {
        state: { offerData, offerSummary, offerFilePath, imported: true },
      });
```

paninaro (changes requested):
Requesting changes to how the coinId is passed from VCCard to VCDetails via useNavigate.

PR #2066, opened 2023-09-08
Move away from lerna to npm with workspaces

paninaro:
Move away from lerna scripts that use package scoping in favor of npm with workspace.
Leaving the build scripts as-is for now.

Some users are reporting issues running the GUI from a source installation on Arch Linux (https://github.com/Chia-Network/chia-blockchain/issues/15934). The issue appears to be with the way `lerna run start --scope @chia-network/gui` would launch Electron, and subsequently, how the GUI launches daemon.


seeden (approved):
lgtm

PR #2067, opened 2023-09-08
Add a tooltip for USDSC (Stably USDS Classic) in the token listing

paninaro:
The token card for Stably USDS Classic now includes a tooltip indicating that the token is the USDSC asset.

Note that this will not rename existing USDS wallets that have been previously added to the wallet DB. As shown below, if the token has not been previously added, adding the token will result in the following:

image

If the token has been previously added (and therefore already has a name), the following is displayed. It is, of course, possible to manually rename the token.
image

The `USDSC` symbol rename is dependent on this PR: https://github.com/Chia-Network/chia-blockchain/pull/16295


hoffmang9 (approved):
lgtm

PR #2069, opened 2023-09-09
Npp remove j2

nirajpathak13:


nirajpathak13:
Note Dependency Review failed because of a known issue with GH internally. This is their response, we can treat it as a false positive:

```
Thanks for contacting GitHub support!

I'm afraid this is a current limitation of Dependency reviews for the GitHub Actions ecosystem. As you'd mentioned, currently for Dependency reviews we do not expand the major version @v2 to the latest version of the action, therefore a Security Advisory whose affected version is denoted as < 2.2.0 will flag @v2 as vulnerable.
​
The Software Engineering team is aware of this issue, and is working on enhancements that will address this scenario, but at this time I'm unable to provide a timeline for when the Software Engineer team will release the aforementioned enhancement.
​
I'd suggest you to keep an eye on the GitHub Blog and GitHub Changelog for future announcements related to Dependency Graph.
​
Let us know if you have any further questions or concerns.
​

Best regards,

Melanie
```

paninaro (dismissed):

PR #2087, opened 2023-09-14
Fixed an issue where expected average passed filter count was wrong

ChiaMineJP:
Fixed the issue: https://github.com/Chia-Network/chia-blockchain/issues/16169


paninaro (approved):
lgtm

PR #2108, opened 2023-09-15
Update messages.po

advlive:


paninaro:
Thanks for the contribution. Updated translations need to be applied to the Crowdin project to be properly integrated into the GitHub repo. If the strings were directly updated in GitHub, they would just get overwritten by the next integration with Crowdin.

advlive:
Hi Actually I am who all persian text made translated, but I saw when apply in gith section as persian start from right side so many items change location exactly when stay in middle english text or ?,!. and this area only can find this problem when happened. So I select some and corrected. not all

advlive:
> Thanks for the contribution. Updated translations need to be applied to the Crowdin project to be properly integrated into the GitHub repo. If the strings were directly updated in GitHub, they would just get overwritten by the next integration with Crowdin.

PR #2121, opened 2023-09-18
fixes for expiring offers

lipalong:
- add view expiration time when viewing offers I created
- fix viewing incoming offer with no expiration time
- fix 'relist offer' option visibility
- color theme fixes


paninaro (approved):
Tested the changes. lgtm

PR #2131, opened 2023-09-19
Refactor out LogoutButton and match style of other header buttons

paninaro:
WalletConnect/Notifications/Logout buttons now have the same dimensions
Logout button has same color as other buttons
Full Node | Wallet button grouping height matches the wallet receive address control
Testnet indicator height matches the wallet receive address control

image


lipalong (approved):
lgtm

PR #2142, opened 2023-09-20
Fix typos

omahs:
Fix typos


paninaro (approved):
Thank you for the corrections @omahs!

PR #2164, opened 2023-09-25
Fix addVCProofs WalletConnect call

freddiecoleman:
### Purpose:

Bugfix of addVCProofs WalletConnect call. Currently it doesn't work because an error is thrown.

### Current Behavior:

- When calling addVCProofs using the HTTP RPC it parses the proofs into a Python dictionary which gets passed into a VCProofs class.
- When calling addVCProofs using WalletConnect it parses the proofs into a string which throws an error when passed into a VCProofs class.

### New Behavior:

- When calling addVCProofs using the HTTP RPC it parses the proofs into a Python dictionary which gets passed into a VCProofs class.
- When calling addVCProofs using WalletConnect it parses the proofs into a Python dictionary which gets passed into a VCProofs class.

### DIFF
--- a/packages/gui/src/constants/WalletConnectCommands.tsx
+++ b/packages/gui/src/constants/WalletConnectCommands.tsx
@@ -879,7 +879,7 @@ const walletConnectCommands: WalletConnectCommand[] = [
     params: [
       {
         name: WalletConnectCommandParamName.PROOFS,
-        type: 'string',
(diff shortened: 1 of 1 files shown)


paninaro (approved):

paninaro:
Merged https://github.com/Chia-Network/chia-blockchain-gui/pull/2166

PR #2177, opened 2023-09-27
Fix crash when viewing a counterable offer

paninaro:
Also fixes localization of "Days", "Hours", "Minutes" in the Setting screen.
Also removes the drop shadow from the emoji picker in dark mode.


packages/gui/src/components/offers2/OfferBuilderViewer.tsx
```diff
@@ -308,7 +304,7 @@ function OfferBuilderViewer(props: OfferBuilderViewerProps, ref: any) {
                 canCounter={canCounterOffer}
```
lipalong, line 304: I think canCounter should be removed here?

paninaro: Thanks, fixed

lipalong (approved):
lgtm

PR #2185, opened 2023-09-29
Skip fee estimator RPC call in wallet mode

paninaro:
Previously it didn't matter whether the full_node process was running or not -- calling useGetFeeEstimateQuery would simply return undefined if the RPC failed because the service wasn't running.

More recently, the service start/stop code was modified such that calling an RPC hook against a service that wasn't running would cause that service to be launched.

This change uses RTK's `skip: <boolean>` option to leave the hook intact (necessary by React's convention to not conditionally call hooks) while signaling that the underlying RPC should not be called when in wallet mode.


wallentx (approved):

emlowe (approved):
I think we should probably take this

Issue #67, opened 2021-02-20
Change "Add Plot Directory" to "Manage Plot Directories"


dvd101x:
On the GUI/Plotting tab on the three dots next to the button " + ADD A PLOT", the second option says "Add Plot Directory", after clicking it, a menu titled "Add a plot" opens with options to "ADD PLOT DIRECTORY" or remove any of the plot directories.

I think some of these issues could be resolved:

1. It could be confusing for someone who would like to remove one or more directories but only finds an option that says "Add Plot Directory", a better description would be "Manage Plot Directories" as referenced on the [FAQ](https://github.com/Chia-Network/chia-blockchain/wiki/FAQ). Even though the FAQ refers to a button but this is a dropdown menu.
2. This menu has an explanation that says "This allows you to add a directory that has plots in it. If you have not created any plots, go to the plotting screen" but doesn't mention what happens if you click on the trashcan icon next to the plot directories.
3. It would be nice for all the buttons that start with the word "ADD" to include a plus sign just like the button " + ADD A PLOT".
4. The trashcan buttons next to each of the plot directories could be changed to a minus sign, to indicate removing instead of deleting (like deleting the directory including the plots) and maybe include an alt text while hovering over them.
5. The menu title is "Add a plot" but this menu is not for adding plots, only adding or removing (managing) plot directories.

hoffmang9:
If there are no plot directories it should say Add a plot directory. If there exist more than 0 plot directories it should say Manage plots.

dvd101x:
Yes. That is so much better better.

Please consider the suggestion about the icons, as the icon to delete plots forever is also a trashcan. And maybe review the title of the dialogue.

dvd101x:
@hoffmang9, sorry to bother you.

Could this be reopened? it is still in the same state as of version 2.0.1
![Add Plot Directory](https://github.com/Chia-Network/chia-blockchain-gui/assets/16261897/5a9cf581-5b46-4815-b957-a68e1f31c26b)

If I want to delete a plot directory, I don't see a button to do it. It feels odd if I want to remove a plot directory to click on add plot directory and then click on the trashcan icon to remove a plot directory.

When I'm in the menu, I see information related to Add plot directories, followed by a list of directories that I can only remove and at the very end it's the option to add.

![Plots](https://github.com/Chia-Network/chia-blockchain-gui/assets/16261897/06b58349-b9b3-4a14-a369-9bb7a30d53d1)

[...]

PR #2199, opened 2023-10-15
Fixed add plot directory button

ChiaMineJP:
Fix this issue: https://github.com/Chia-Network/chia-blockchain-gui/issues/67

Changed icon and caption
(Before this PR, it is always `Add Plot Directory` even if an user want to remove plot directory from this menu)
image

Added info level alert and tooltip
image


seeden (approved):
lgtm

Issue #2181, opened 2023-09-28
Requested NFTs shouldn't list owned NFTs [Bug] 


danieljperry:
### What happened?

When creating an offer to request an NFT, there is a drop-down that lists all NFTs this wallet owns. I don't think there is a reason to request NFTs you already own, so this drop-down seems superfluous.

In the following image, the wallet already owns #9079 and #2365:

![requestnfts](https://github.com/Chia-Network/chia-blockchain-gui/assets/69756004/f94c2f70-cd8e-4b91-9785-09ec454bfd60)

### Version

2.1.0rc3 (binary installation)

### What platform are you using?

Windows

### What ui mode are you using?

GUI

PR #2204, opened 2023-10-17
Don't show NFT collection options on the request side of the offer builder

paninaro:
Addresses https://github.com/Chia-Network/chia-blockchain-gui/issues/2181

When using Offer Builder to construct an offer, the NFT request column will no longer provide autocomplete options from the user's own NFT collection. The offer column will continue to populate the NFT autocomplete options as before.


seeden (approved):
lgtm

Issue #2210, opened 2023-10-22
[Bug] Disabling expiration time is not persisted


Ealrann:
### What happened?

Hello,

I want to disable the "Offer Expiration Time" on my linux wallet, so I go into the settings and click on "Disable" (btw, I have to click 2 times on the Disable button before it works).

It works for the current session, but every time I restart the application, I have to disable it again. 

![image](https://github.com/Chia-Network/chia-blockchain-gui/assets/723000/3c5fbd8c-3bad-4ab5-80ce-bc8f5b5b539f)

### Version

2.1.1

### What platform are you using?

Linux

### What ui mode are you using?

GUI

PR #2214, opened 2023-10-25
Fix accidental mutation of the expiring offers defaults object

paninaro:
Fixes https://github.com/Chia-Network/chia-blockchain-gui/issues/2210


packages/gui/src/hooks/useOfferExpirationDefaultTime.tsx
```diff
@@ -47,15 +47,13 @@ function stringPropertiesToNumbers(objWithStrings: OfferExpirationDefaultTimeInp
 }
 
 export default function useOfferExpirationDefaultTime() {
+  const defaultsCopy = { ...offerExpirationDefaultTimeDefaults };
```
paninaro, line 50: We were returning the `offerExpirationDefaultTimeDefaults` object, allowing direct write access to the object. This caused issues with the logic for updating the persisted pref values.

seeden (approved):

Issue #2182, opened 2023-09-28
Delete notification requires double confirmation [Bug] 


danieljperry:
### What happened?

When deleting an offer notification, two similar confirmation dialogs are displayed, as shown in this video:

https://github.com/Chia-Network/chia-blockchain-gui/assets/69756004/dbe78d06-28b5-44a0-acca-bd79de8161a9

### Version

2.1.0rc3 (binary installation)

### What platform are you using?

Windows

### What ui mode are you using?

GUI

PR #2215, opened 2023-10-26
Remove a redundant confirmation when deleting an offer notification

paninaro:
Fixes https://github.com/Chia-Network/chia-blockchain-gui/issues/2182

### DIFF
--- a/packages/gui/src/components/offers2/OfferIncomingTable.tsx
+++ b/packages/gui/src/components/offers2/OfferIncomingTable.tsx
@@ -1,16 +1,5 @@
 import { toBech32m } from '@chia-network/api';
-import {
-  Card,
-  Flex,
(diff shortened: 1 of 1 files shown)


packages/gui/src/components/offers2/OfferIncomingTable.tsx
```diff
@@ -211,17 +199,7 @@ export default function OfferIncomingTable(props: OfferIncomingTableProps) {
 
   async function handleDeleteNotification(notification: Notification) {
     try {
-      const canProcess = await openDialog(
```
paninaro, line 214: The `deleteNotification` code path displays its own confirmation dialog, so this confirmation is redundant.

seeden (approved):

PR #2221, opened 2023-11-02
added missing change directory command before running npm run dev

RightSexyOrc:


paninaro (approved):
Thank you for the correction!

PR #2231, opened 2023-11-22
Updated validTimeList assignment for null safety in Offer Summary

grigb:
This commit enhances the reliability of the validTimeList assignment. Previously, the code directly accessed offerSummary.validTimes, which lead to issues if offerSummary was undefined. This created an error when displaying offers in the Chia GUI.

The error breaks the GUI, and the user is prompted to reload the app.

The updated code now safely checks for the existence of offerSummary before attempting to access its validTimes property. If offerSummary is null or undefined, validTimeList is explicitly set to null, thereby preventing any undefined reference errors.


grigb:
Fix was not needed. It was already implemented.

PR #2249, opened 2023-12-24
Walletconnect create offer fee

Ganbin:
- Added the ability to add a fee when calling Wallet Connect `createOfferForIds`.
- The fee is optional.
- If the fee is not given, the user still has the ability to add a fee in the created offer.

## Todo
- [x] Add and adapt unit tests


paninaro (approved):
Looks good to me. I tested the change and also verified that offer creation through normal non-WalletConnect methods works as expected. Thanks for the change and thank you for adding some tests.

PR #2281, opened 2024-02-23
checkpoint: into main from release/2.2.0  @ abb10ab4405dc0f845b912c690594458284ae21e

AmineKhaldi:
Source hash: abb10ab4405dc0f845b912c690594458284ae21e
Remaining commits: 3

### DIFF
--- a/packages/gui/src/util/chiaEnvironment.js
+++ b/packages/gui/src/util/chiaEnvironment.js
@@ -70,14 +70,27 @@ const getChiaVersion = () => {
 const startChiaDaemon = () => {
   const script = getScriptPath(PY_DIST_FILE);
   const processOptions = {};
-  // processOptions.detached = true;
-  // processOptions.stdio = "ignore";
+  // We want to detach child daemon process from parent GUI process.
+  // You may think `detached: true` will do but it shows blank terminal on Windows.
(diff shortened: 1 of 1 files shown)


emlowe:
There was some problem with these changes in the release branch and I'm not sure we want to checkpoint them back into main - checking with @ChiaMineJP

AmineKhaldi:
> There was some problem with these changes in the release branch and I'm not sure we want to checkpoint them back into main - checking with @ChiaMineJP

We'll end up naturally checkpointing 82168f1 which gets us where we want to be. Am I missing something?

altendky:
My understanding was that we would checkpoint everything but no-op the removal in https://github.com/Chia-Network/chia-blockchain-gui/commit/82168f1b10907a3ff379a449106f4071c80c4e0c.  Then release does not have the feature and main has the feature, bugs and all, and it can be fixed there.

As a detail, _everything_ gets checkpointed.  The only variation is how you do it.  Like making the diff empty when we checkpoint https://github.com/Chia-Network/chia-blockchain-gui/commit/82168f1b10907a3ff379a449106f4071c80c4e0c since that change (removal) is meant only for the release branch.  But we have to record that intent in git.

altendky (approved):

PR #2308, opened 2024-03-18
Adding Walletconnect endpoints for DIDs, Datalayer, and NFT Bulk Mint

Gaerax:
Added a walletconnect endpoints for almost all DID and Datalayer commands, and nft_mint_bulk and push_tx to enable bulk minting via WC, and creating feature rich dapps.


Gaerax:
I also added an endpoint to allows an app to request permissions for a list commands. This would make for a much better user experience for dapps that require use of many commands, as the user can just be prompted once, at app launch to give permissions for all of the basic, read-only commands that it needs.

This includes a new confirm dialog, which could probably be improved, but it does the job.
image

If the app already has all of the permissions in the list, the command returns successful without prompting the user.

If the command list sent by the app contains a command that does not allow bypass, or is not a valid command name, it is ignored.

Gaerax:
Since updating my repo with recent commits for the PR, I have found that some of the new WC commands which used to work fine, now have issues. I will switch this PR to a draft for the time being, until I figure out why that is.

Gaerax:
There was no issue, I'm just a dumb dumb who forgot to run the build command

ChiaMineJP:
Thank you for the great PR!
Please give us time to review. We'll check the code.

packages/api-react/src/services/wallet.ts
```diff
@@ -1132,7 +1132,27 @@ export const walletApi = apiWithTag.injectEndpoints({
...
+
+    // findLostDID: mutation(build, DID, 'findLostDID', {
+    //   invalidatesTags: (_result, _error, { coinId }) => [
+    //     { type: 'DIDInfo', id: coinId },
+    //     { type: 'DIDCoinInfo', id: coinId },
+    //   ],
+    // }),
+
+    getDIDMetadata: query(build, DID, 'getDidMetadata', {
+      providesTags: (result, _error, { walletId }) => (result ? [{ type: 'DIDMetadata', id: walletId }] : []),
```
ChiaMineJP, line 1146: The tag name `DIDMetadata` should be added to line#10 `tagTypes` array.

packages/api-react/src/services/wallet.ts
```diff
@@ -1132,7 +1132,27 @@ export const walletApi = apiWithTag.injectEndpoints({
...
+    }),
+
+    updateDIDMetadata: mutation(build, DID, 'updateDidMetadata', {
+      invalidatesTags: (_result, _error, { walletId }) => [
+        { type: 'DIDInfo', id: walletId },
+        { type: 'DIDCoinInfo', id: walletId },
+        { type: 'Wallets', id: walletId },
+        { type: 'DIDWallet', id: walletId },
```
ChiaMineJP, line 1154: `{ type: 'DIDMetadata', id: walletId}` should be added.

ChiaMineJP (changes requested):
I've just finished a part of reviewing. I'll do more check later.

emlowe:
Closing PR - @Gaerax if you want to respond to the code change requests you can do so and reopen the PR

PR #2313, opened 2024-03-24
pnpm poc

wallentx:
Drafting monorepo management via pnmp - yarn v4 monorepo management PR coming shortly for comparison


wallentx:
To provide a better description of what I'm exploring in this PR:

1. In November 2023, node started shipping with `corepack`, as well as `pnpm`, and `yarn` (not enabled by default, but `corepack enable` enables them), and they also claim they have no intentions to replace `npm` as the default package manager. Both `pnpm` and `yarn` support management of monorepos by default, so you could ditch `lerna` and `nx`. Though, @seeden said he didn't want to shift away from `npm` at the moment.

2. I was looking at build+install speed differences between using `pnpm`, `yarn`, and `npm`. I also saw that the webpack usage - the longest running task during gui install - might be able to leverage `thread-loader` for a speedup.

3. I was always bothered by the install time of the GUI, so I added a `build:local` script to only generate locales that are on my system, and to also drop usage of `cross-env` for when building locally.

4. It was Sunday and I wanted to learn a bit about new nodejs stuff, so I used this as an excuse, and also, my GitHub activity is full of crickets, so I felt anxious and needed to make some commits.

This PR can be ignored for the moment, since @seeden said he wants to stick with NPM, _however_, my findings from these changes when building on my phone did show a really good speedup.
`main`: DNF. I got impatient and stopped the install at 27 minutes 
`wallentx/pnpm`: 19 min

So I might transform this PR to just the speedups if I can replicate them using `npm`. But at the moment, I'm actually having issues with the webpack config on x86_64 for some reason

seeden:
@wallentx, thank you for your PR. I understand that pnpm is faster, but we are using monorepo with a flat structure. pnpm creates a hierarchical structure, which is the main reason why you see many webpack issues during the build. Because of this, you need to be very careful with all package versions. You can easily end up with two different versions in your final bundle, which brings "crazy" unexpected errors across the application. 
We will save 10 minutes, but the development effort will take you a few other hours of your life fixing problems with versions.

wallentx:
Understood, and thanks for the explanation! I'll go ahead and close this, and if I can achieve any speedup just using `thread-loader`, I'll make a separate PR for that.

PR #2336, opened 2024-05-07
graph performance

seeden:
fix line graph performance


ChiaMineJP:
@seeden looking into. Please give me time.

packages/core/src/components/LineChart/LineChart.tsx
```diff
@@ -0,0 +1,117 @@
...
+  const {
+    data,
+    // min: defaultMin = 0,
+    xValueFormatter = defaultXValueFormatter,
+    yValueFormatter = defaultYValueFormatter,
+    height = 150,
+  } = props;
+
+  const stringifiedData = useMemo(() => JSONbig.stringify(data), [data]);
+  const freezedData = useMemo<Point[]>(() => JSONbig.parse(stringifiedData), [stringifiedData]);
```
ChiaMineJP, line 69: Out of curiosity, why stringified then parse?
Was it for debug?
Since it seems there are no array/object manipulation, I think you can directly use `data` to get `yData` and `xData` without any side effects

packages/core/src/components/LineChart/LineChart.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+  const yData = useMemo(() => freezedData.map((item) => item.y), [freezedData]);
+  const xData = useMemo(() => freezedData.map((item) => item.x), [freezedData]);
+
+  const yDataNumber = useMemo(
+    () => yData.map((value) => (value instanceof BigNumber ? value.toNumber() : value)),
+    [yData],
+  );
```
ChiaMineJP, line 77: Not a big deal.
I think summarizing the use of `React.useMemo` can slightly reduce the CPU/RAM consumption because the number of  `[deps]` to remember / check diff is reduced.

So I'd suggest to remove `yData` lines and integrate it to code for `yDataNumber`
```typescript
const yDataNumber = useMemo(
  () => data.map((item) => item.y instanceof BigNumber ? item.y.toNumber() : item.y),
  [data],
);
```

packages/core/src/components/LineChart/LineChart.tsx
```diff
@@ -0,0 +1,117 @@
...
+  // const min = data.length ? Math.min(...yDataNumber) : defaultMin;
+  // const max = Math.max(min, ...yDataNumber);
+
+  const xAxis = useMemo(
+    () => ({
+      data: xData,
+      valueFormatter: xValueFormatter,
+    }),
+    [xData, xValueFormatter],
+  );
```
ChiaMineJP, line 88: Similar to above.
Since It seems `xData` is only an intermediate variable and not be used after getting the `xAxis`, the code to get `xData` can be summarized into the code for `xAxis`.

packages/wallets/src/components/WalletGraph.tsx
```diff
@@ -173,63 +132,37 @@ export default function WalletGraph(props: WalletGraphProps) {
...
-  });
+  const data = prepareGraphPoints(balance, confirmedTransactions);
+
+  const xValueFormatter = useCallback((value: number) => moment(value * 1000).format('LLL'), []);
 
-  const min = data.length ? Math.min(...data.map((item) => item.y)) : 0;
-  const max = Math.max(min, ...data.map((item) => item.y));
+  const yValueFormatter = useCallback(
+    (value) => {
+      const formatedValue = isCAT ? mojoToCAT(value) : mojoToChia(value);
```
ChiaMineJP, line 148: Typo: `formatedValue` -> `formattedValue`

packages/wallets/src/components/WalletGraph.tsx
```diff
@@ -173,63 +132,37 @@ export default function WalletGraph(props: WalletGraphProps) {
...
-    offset: 0,
-  });
+  const data = prepareGraphPoints(balance, confirmedTransactions);
+
+  const xValueFormatter = useCallback((value: number) => moment(value * 1000).format('LLL'), []);
 
-  const min = data.length ? Math.min(...data.map((item) => item.y)) : 0;
-  const max = Math.max(min, ...data.map((item) => item.y));
+  const yValueFormatter = useCallback(
+    (value) => {
```
ChiaMineJP, line 147: My VSCode complains a missing type.
According to the `yValueFormatter` definition by `LineChart`, this should be `(value: number | BigNumber | null) => {`

ChiaMineJP (changes requested):

packages/core/package.json
```diff
@@ -75,6 +75,7 @@
     "@mui/material": "5.14.5",
     "@mui/styled-engine-sc": "5.12.0",
     "@mui/styles": "5.14.5",
+    "@mui/x-charts": "7.3.2",
```
markelrod, line 78: GItHub is warning that '@mui/x-charts' should be listed in the project's dependencies, not devDependencies

markelrod (approved):

ChiaMineJP (approved):

PR #2360, opened 2024-05-26
Bump lerna to 8.1.3, ejs to 3.1.10

ChiaMineJP:
This PR looks like just updating `lerna` version but it does not such simple.

## Steps to reproduce this change

1. Check out `main` branch
2. Delete the root `package-lock.json` and `node_modules`. 
3. Run `npm cache clean -f`
4. Update the root `package.json` like below then run `npm i`
  - Update `lerna` version to "8.1.3"
  - Update `nx` version to `19.0.1` 
  - Add the following section to `package.json`. 
  ```
  "overrides": {
    "@walletconnect/keyvaluestorage": "1.0.2",
    "tslib": "2.6.2"
  },
  ```

## Why such complicated?
My environment: m1 Mac 14.3.1, NodeJS 18.20.3, npm 10.8.0

At first I just incremented lerna version to `8.1.3`. From that point, `npm run build` keeps failing with the strange message 

image

After some investigation, I found that the version of `node_modules/tar/node_modules/minipass` was incorrectly set to `3.3.6` where `tar` specifies the version  to `^5.0.0`.

image

image

I suspect that with workspace and multiple child package dependencies, `npm` was confused and set wrong versions when creating `package-lock.json`.

### Challenge 1
1. check out `main`
2. Delete the root `node_modules`
3. Set `"lerna": "8.1.3",` in the root `package.json`
4. Run `npm i`
5. Run `npm run build` <-- This fails

### Challenge 2
1. check out `main`
2. Delete the root `node_modules` and `package-lock.json`
3. In the root `package.json`,  set `"lerna": "8.1.3"`, remove `"nx": "16.7.2"`
4. Run `npm i`
5. Run `npm run build` <-- This succeeds
6. Run `npx lerna run dev:skipLocales --scope=@chia-network/gui` to start GUI <-- GUI crashes with an error `(0 , tslib__WEBPACK_IMPORTED_MODULE_4__.__spreadArray) is not a function`

## Lessons learned
- Even if the content of `package.json` is the same, the content of `package-lock.json` differs according to previous `package-lock.json` and `node_modules` content.
- `lerna` depends on `nx` and `nx` has several optional dependencies for various OSes. This messes up `package-lock.json`. When I removed `nx` from top-level dependencies in the root `package.json` and ran `npm i`, it generated `package-lock.json` which specifies one of the optional dependencies only for MacOS.  Of cource, this `package-lock.json` fails except for macOS.


ChiaMineJP:
I tested the following items and confirmed all ok.
- `npm i`
- `npm run build`
- `npx lerna run dev:skipLocales --scope=@chia-network/gui` (Launch GUI)
- Connect sample dApp with walletconnect, tested some APIs then disconnect.

markelrod (approved):
Tested on x86 MacOS 14.4.1

PR #2403, opened 2024-06-15
Keep latest_coin_id in GUI NFT signature response so it is same as RPC response

markelrod:


ChiaMineJP (approved):
I searched for the reason why `latest_coin_id` was removed in the past but found no clue.
My guess is that `latest_coin_id` isn't referred by the signature verification process so the author might feel it's useless.

emlowe (approved):

PR #2404, opened 2024-06-15
Prevent the Enter key from submitting the form when text input has focus

markelrod:
- Prevent the Enter key from submitting the form when text input has focus
- Include our NFT collection in option list for signing


ChiaMineJP (approved):
Tested in my m1 mac environment. Works fine!

emlowe (approved):

PR #2410, opened 2024-06-18
update pool points info for filter change

BrandtH22:


markelrod (approved):

ChiaMineJP (commented):
There is a config for soft fork heights for plot filter update.
https://github.com/Chia-Network/chia-blockchain-gui/blob/main/packages/api/src/constants/PlotFilter.ts

If the point values in the text in this PR are variable and determined by current plot filter , it will be more accurate.

And this PR (https://github.com/Chia-Network/chia-blockchain-gui/pull/2406) should be merged first if you want to introduce variable text according to plot filter value.

PR #2446, opened 2024-08-14
Hide to-address in wallet transaction history

ChiaMineJP:
image

image


markelrod:
Hey Izumi - we're getting feedback that hiding the To-address is going to be seen as a bug. For now let's truncate the address for display and not allow user to copy the full address. Does that make sense?

ChiaMineJP:
@markelrod Updated the code to do so.

https://github.com/user-attachments/assets/e19ee9d2-fb1c-4bb7-ac58-3f2e5787c7d7

markelrod (approved):

PR #2526, opened 2024-11-04
Remove unused code that attempts to read the output of the daemon

emlowe:
This old code was rendered unused by #2309 - but probably has not been used since earlier than that. Once the GUI starting launching the daemon via `chia start daemon` it's not able to read the stdout as the process it launches is the chia CLI which then launches the daemon, so the GUI is one process removed from being able to read this.

The daemon currently does output on stdout some JSON that was used by this code at some point in the past. That JSON code is no longer needed and is being removed in a chia-blockchain PR https://github.com/Chia-Network/chia-blockchain/pull/18809 

This PR doesn't need to wait for that to be merged, as the current GUI doesn't read this anyway.


ChiaMineJP (approved):
lgtm

aqk (approved):

markelrod (approved):

PR #2609, opened 2025-02-18
Fix sync status

emlowe:
Made the behavior of FullNode sync indicator aligned with that of the CLI


emlowe:
close and reopen for CI

wjblanke (approved):
aok

PR #2611, opened 2025-02-19
use nft id as filename when the user downloads multiple nfts at once

cmmarslender:
simplify logic for downloading multiple files
close files during error


emlowe (approved):
Looks good

arvidn (approved):

PR #2734, opened 2025-08-24
Support CAIP-25 style session proposal handling

ChiaMineJP:
https://specs.walletconnect.com/2.0/specs/clients/sign/session-proposal
https://reown.com/blog/caip-25-implementation-guidance-migrating-to-empty-undefined-required-namespaces

### Why this change
WalletConnect is migrating to align with the CAIP-25 standard, which treats all namespace requests as optional for better UX.

#### Old pattern
dApps use `requiredNamespaces` (must support or connection fails)
#### New pattern
dApps use `optionalNamespaces` only, leaving `requiredNamespaces` empty/undefined

This allows wallets to approve connections even when they don't support all requested chains

This PR still supports `requiredNamespaces`, because many dApps haven't migrated yet, so we need backward compatibility.

### DIFF
--- a/packages/gui/src/util/walletConnect.ts
+++ b/packages/gui/src/util/walletConnect.ts
@@ -44,7 +44,13 @@ export async function processSessionProposal(
           icons?: string[];
         };
       };
-      requiredNamespaces: {
+      requiredNamespaces?: {
+        chia: {
+          chains: string[];
+          methods: string[];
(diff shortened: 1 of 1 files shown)


seeden (approved):

ChiaMineJP:
Close because #2842 superseeded this PR

PR #2906, opened 2026-04-24
We use lerna. lerna uses Nx. Nx has a bug that causes lerna to hang. Disable Nx.

aqk:
We use lerna. lerna uses Nx. Nx has a bug that causes lerna to hang. Disable Nx.

---

> [!NOTE]
> **Low Risk**
> Low risk config-only change that alters Lerna’s task runner behavior; primary risk is slower runs or differences in caching/parallelism when Nx is disabled.
> 
> **Overview**
> Disables Nx integration for Lerna by setting `useNx: false` in `lerna.json`, avoiding Nx-related hangs during monorepo commands.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 657e30590d9858e35e0840dbfa56b82159fdeb76. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,4 +1,5 @@
 {
   "packages": ["packages/*"],
-  "version": "1.0.0"
+  "version": "1.0.0",
+  "useNx": false
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -16396,9 +16396,9 @@
       "license": "ISC"
     },
     "node_modules/follow-redirects": {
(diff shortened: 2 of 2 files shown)


aqk:
Unfortunately, this breaks CI. We can use this workaround locally, without changing how CI works.

PR #2970, opened 2026-07-07
[CHIA-4294] Fixed electron source install issue

ChiaMineJP:
## Summary

Fixes Electron failing to install on Node.js **24.16.0+** with:

```
Error: Electron failed to install correctly, please delete node_modules/electron and try installing again
```

### Root cause

Node 24.16.0+ shipped a zlib change that breaks **yauzl 2.x**'s unzip path. Electron's install downloads a zip and extracts it via `@electron/get` → `extract-zip` → `yauzl@^2.10.0`, so on affected Node versions extraction silently fails — `node_modules/electron` ends up with no usable binary (`path.txt` missing / incomplete `dist/`).

See [electron/electron#51619](https://github.com/electron/electron/issues/51619).

### Fix

Force `yauzl@3.4.0` via the root `package.json` **`overrides`** block (and regenerate the lockfile entry for `node_modules/yauzl`):

```json
"overrides": {
  "yauzl": "3.4.0",
  ...
}
```

### Why `overrides`, not a top-level `dependencies` entry

`extract-zip` hard-requires `yauzl@^2.10.0`. Adding `yauzl@3.4.0` as a sibling top-level dependency does **not** replace the nested transitive copy that `extract-zip` resolves — npm still installs 2.10.0 for that path. `overrides` is required so the nested copy under `extract-zip` is forced to 3.4.0 as well.

After install, `npm ls yauzl` should show:

```
extract-zip@2.0.1 └── yauzl@3.4.0
```

(npm may report `invalid: "^2.10.0"` because of extract-zip's range; that is expected with an override.)

### Related

Companion belt-and-suspenders guards in the parent repo (`install-gui.sh` / `Install-gui.ps1`) re-run `node node_modules/electron/install.js` when `path.txt` is missing: [Chia-Network/chia-blockchain#21105](https://github.com/Chia-Network/chia-blockchain/pull/21105). Those alone are insufficient on Node 24.16+ while yauzl 2.x is still in use; this override is the real fix.

## Test plan

- [ ] On Node 24.16.0+ (or current Node 24.x): `rm -rf node_modules && npm ci`
- [ ] Confirm `npm ls yauzl` shows `yauzl@3.4.0` under `extract-zip`
- [ ] Confirm `node_modules/electron/path.txt` exists and the Electron binary runs (`electron --version` / platform-equivalent path)
- [ ] Confirm `npm run electron` starts without the “failed to install correctly” error

---

> [!NOTE]
> **Low Risk**
> Dependency-only override affecting dev-time Electron install tooling; no runtime app logic changes.
> 
> **Overview**
> Adds **`yauzl@3.4.0`** to the root `package.json` **`overrides`** block and updates the lockfile so the transitive copy used by **`extract-zip`** (Electron’s zip download path) is no longer **2.x**.
> 
> On **Node.js 24.16.0+**, zlib behavior breaks **yauzl 2.x** unzip, so Electron’s postinstall can fail silently and leave an incomplete **`node_modules/electron`** install. Forcing **3.4.0** restores reliable extraction without changing application source.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 31fe1b7990942dcdb13994d9c2ceb3128b7c787f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


seeden (approved):

emlowe:
This isn't needed anymore as the move to electron 43.4.0 removes the extract-zip dependency and by side effect also fixes this - see https://github.com/electron/electron/issues/51619#issuecomment-4682930068

PR #2981, opened 2026-07-14
[CHIA-4310] Added monkey patching to properly parse bigints

matt-o-how:
WalletConnect needed patching for parsing bigints

---

> [!NOTE]
> **Medium Risk**
> Changes run on every `npm install` and mutate WalletConnect dependencies used in wallet/session flows; scope is narrow but incorrect patching could break JSON parsing for offers and amounts.
> 
> **Overview**
> Fixes WalletConnect JSON parsing so **negative** mojos encoded as strings like `-100n` become real `BigInt` values instead of staying as strings (which broke `parseMojos` on offer amounts).
> 
> Adds **`patch-package`** with a checked-in patch on `@walletconnect/safe-json` that widens the bigint detector from `/^\d+n$/` to `/^-?\d+n$/` in CJS, ESM, and UMD builds. **`postinstall`** now runs `patch-package`, then **`scripts/fix-walletconnect-bigint-regex.js`**, which scans other `@walletconnect` `.js` bundles that inline the same regex and applies the same replacement. The fix script is listed in **`.eslintignore`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f7cd5156387fd4e6bd164e6584a760319610dfe5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


seeden (dismissed):

seeden (dismissed):

seeden (approved):

wjblanke (approved):
aok

Rigidity (commented):
I'm not sure why this was merged, but it should be reverted. Patching WalletConnect is a hacky and fragile workaround for large number serialization issues on the client side. dApps can (and generally do) encode numbers larger than 2^53-1 (the JS safe integer limit) as strings. The wallet should in turn support decoding these and do the same for its responses.

PR #2681, opened 2025-06-18
update electron and isolate api

seeden:
update electron to latest version
simplify and isolate api
add double confirmation for send transaction and other commands

PR #2993, opened 2026-07-22
Fix NFT video/audio playback, persist cache settings, add video looping, and improve NFT gallery UX

jlobue10:
## Summary

Video and audio NFTs frequently rendered as blank tiles and could not be played anywhere in the GUI. This PR fixes the media pipeline end to end and addresses several related gallery/settings issues found along the way.

### 1. Fix NFT video/audio playback (media pipeline)

- Register the `cache:` scheme with `registerSchemesAsPrivileged` (`stream: true`, plus `standard`/`secure`/`supportFetchAPI`). Without this, media elements expect a buffered response and stall on the streamed body served by `protocol.handle`.
- Serve HTTP 206 partial content for Range requests from the cache protocol handler. Media elements seek via Range requests; without 206 responses seeking was broken and MP4 files without faststart (moov atom at the end) could fail to start playing at all.
- Make the 30s download timeout an inactivity timeout (reset on each received chunk) instead of a total-transfer budget, so larger videos on slower hosts can finish downloading while stalled connections still fail fast.
- Report a distinct "Maximum file size exceeded" error when the size cap aborts a download, persist it, and only retry when the caller lifts the size limit — previously these surfaced as generic transient abort errors and oversized files were re-downloaded on every gallery visit.
- `maxSize <= 0` now disables the size limit instead of aborting every download on the first chunk (fixes the `ignoreSizeLimit` path; `NFTPreview` forwards the same override so verification and display stay consistent).
- Drive-by: download progress used `Math.max` instead of `Math.min` and was pinned at 100%.

### 2. Fix unclickable media controls in the NFT detail view

The electron isolation refactor (#2681) inverted a condition to `allowPointerEvents={!canInteract}`, disabling pointer events on the sandboxed iframe exactly when interaction should be allowed — so the play button in the detail view never received clicks.

### 3. Persist NFT cache size and folder across GUI restarts

The settings UI stored the cache size limit under the `cacheLimitSize` preference key, but startup reads `maxCacheSize` — a key nothing ever wrote — so the setting silently reverted to the 1GB default on every launch, and a changed cache folder was never persisted at all. Both are now persisted from `CacheProvider` when change events arrive from the main process (the renderer is the only safe writer since `prefs.yaml` is rewritten from the renderer's full snapshot on every preference save). The legacy `cacheLimitSize` key is still honored at startup, and non-positive stored values no longer crash the `CacheManager` constructor.

### 4. Gallery: reliable updates, refresh button, inline playback

- Newly acquired NFTs could take a long time to appear (or required switching wallets and back). The `nft_coin_added` subscription was torn down and recreated on every `NFTProvider` render because `useNFTCoinEvents` passed fresh inline callbacks each render — wallet events arriving in the resubscribe gap were silently lost. The callbacks are now stable.
- The NFT data provider exposes `refetch()` and the gallery header gains a Refresh button as a manual fallback.
- Video/audio NFTs are playable directly in the gallery grid: pointer events are enabled on playable tiles so the native media controls respond with a single click, while transparent blockers over the non-control area keep clicks there navigating to the detail view as before.

### 5. NFT video looping (global and per-video)

- Settings > NFT gains a "Loop videos" switch that forces looping for all NFT videos.
- Each video player (gallery grid, detail view, fullscreen) gets a loop toggle button overlaid on the top-right corner of the player. The native video-controls menu is Chromium-internal and not extensible, and the sandboxed iframe blocks scripts, so an overlay outside the iframe is the only viable placement. Toggling remounts the iframe (the video restarts paused) — unavoidable without weakening the sandbox.
- Preferences persist as a global flag plus a per-NFT map.

### 6. Show NFT previews without blocking on full-file hash verification (offer viewer)

Previews were gated behind `useNFTVerifyHash` completing, which downloads and checksums the complete data file of every URI. In the offer viewer the NFTs involved are typically not in the local cache, so every preview sat on an endless spinner (minutes for large videos) or turned into a blank tile with a misleading "Invalid hash" chip when the download failed (dead host, network error, or file over the cache size cap). The gallery masked the problem because owned NFTs are usually already cached.

- `useNFTVerifyHash` now exposes the most likely preview URI immediately (metadata preview video → preview image → data URI) while checksums run in the background, skipping candidates whose validation has already failed so a fast failure on one source doesn't flash an error tile while another is still pending.
- Download failures are flagged separately from real hash mismatches (`failedFetch`), and a hash mismatch on any URI outranks a download failure so a tampered file is never reported as merely unavailable.
- `NFTPreview` renders as soon as the preview file itself is fetched, shows an honest "Loading preview…" while it downloads, and an explicit "Preview is not available" state instead of a silent blank iframe when the file cannot be retrieved. `NFTHashStatus` reports "File is not available" for download failures; "Invalid hash" remains reserved for genuine mismatches.
- Videos with a metadata preview image now show their thumbnail immediately instead of waiting for the full video to download.

### 7. Fix broken NFT previews in the dapp confirmation dialog

The take-offer confirmation dialog (shown when a paired dapp such as dexie requests `chia_wallet.take_offer`) renders NFT previews in its You Spend / You Receive sections with an `` pointed at the NFT's first data URI. For video NFTs that is a video file, which an `` cannot display — every video NFT appeared as a broken image icon. The preview URL is now resolved file-type-aware: an image data URI is used directly; otherwise the NFT metadata is fetched (best effort, bounded size/timeout) and its `preview_image_uris` used; extensionless data URIs (common on IPFS) are still attempted; NFTs with no usable image fall back to the existing placeholder tile. Applies to both sides of the dialog.

## Bug audit

The full branch diff was put through three independent adversarial code reviews (media pipeline / state & subscriptions / preview rendering), with every candidate finding verified against the code before being accepted. Four real issues were found and are fixed in this PR:

1. **Premature fallback to a failed data source during verification** (`useNFTVerifyHash`): while checksums ran in parallel, whichever source settled first won — a fast failure on the data URI could flash "Preview is not available" while a perfectly valid preview image was still being verified, then snap to the correct preview. The candidate selection now skips already-failed sources and only reports a failure once verification has finished.
2. **Download failure masking a hash mismatch** (`useNFTVerifyHash`): with multiple data URIs, a fetch failure on the first URI hid a checksum mismatch on a later one behind the benign "File is not available" label. Mismatches now take precedence.
3. **Loop button vs. selection checkbox** (`NFTCard`): in gallery multi-select mode the new loop toggle rendered on top of the selection checkbox and swallowed its clicks. Media interactions (and the loop button) are now disabled while selecting, so the whole tile selects on click.
4. **Unbounded download duration** (`downloadFile`): the inactivity timeout alone let a host trickling bytes hold one of the ten download concurrency slots indefinitely. Downloads now also have an absolute 30-minute deadline, reported as the same transient abort error so legitimate retries still work.

The reviews also specifically cleared: Range-request edge cases (open-ended/suffix/out-of-bounds/zero-length), concurrent same-URL download dedup and partial-file safety, cache prefs persistence ordering (no stale-value or write-loop paths), refresh/refetch abort safety on double-trigger, `nft_coin_added` subscription lifecycle, iframe remount semantics (same-URI transitions don't restart playback; content changes do), and the legacy `cacheLimitSize` migration path.

## Testing

- Automated Electron harness exercising the real `CacheManager` protocol handler: correct 200/206/416/404 responses with byte-exact range bodies, and real `<video>`/`<audio>` playback + seeking through `cache://` for faststart MP4, non-faststart MP4 (moov at end), and AAC audio — both top-level and inside the production-CSP sandboxed iframe. A control run against the pre-fix code reproduces the failures (full-file 200s, stalled media elements).
- Automated Electron harness rendering the real `NFTCard`/`NFTPreview`/`NFTProvider`/`OfferBuilderViewer` tree with live mainnet NFT data and a live offer summary, across cold-cache, slow-download, and failed-download scenarios — verifying immediate previews, honest loading states, and the explicit failure state.
- The real dapp confirmation dialog (`openReactDialog` + `Confirm` + `parseCommandDisplay`) rendered against a live daemon and a live NFT offer: picture NFTs show thumbnails, video NFTs show the metadata preview image when available and the placeholder (not a broken image) when not.
- 278 jest tests pass (3 new for the dialog preview resolution); eslint/prettier clean; tsc error baseline unchanged.
- Manual testing on Linux with real NFTs: gallery previews, detail-view playback with seeking, inline grid playback, cache size/folder persistence across restarts, refresh button, video looping, offer viewer, and the dexie take-offer confirmation dialog.

### DIFF
--- a/packages/gui/src/components/cache/CacheProvider.tsx
+++ b/packages/gui/src/components/cache/CacheProvider.tsx
@@ -1,3 +1,4 @@
+import { usePrefs } from '@chia-network/api-react';
 import React, { useMemo, useState, useEffect, useCallback, type ReactNode } from 'react';
 
 const { cacheAPI } = window;
@@ -15,6 +16,14 @@ export default function CacheProvider(props: CacheProviderProps) {
   const [cacheDirectory, setCacheDirectory] = useState<string | undefined>(undefined);
   const [cacheSize, setCacheSize] = useState<number | undefined>(undefined);
 
+  // The main process only reads these preferences at startup; changes made at
+  // runtime live in CacheManager's memory, so they are persisted here in the
+  // renderer where all other preferences are written (prefs.yaml is rewritten
+  // from the renderer's snapshot on every preference save - a main process
+  // write would be clobbered by the next renderer save).
+  const [, setMaxCacheSizePref] = usePrefs<number | undefined>('maxCacheSize', undefined);
+  const [, setCacheFolderPref] = usePrefs<string | undefined>('cacheFolder', undefined);
+
   const updateCacheSize = useCallback(async () => {
     const size = await cacheAPI.getCacheSize();
     setCacheSize(size);
@@ -30,9 +39,21 @@ export default function CacheProvider(props: CacheProviderProps) {
     setMaxCacheSize(size);
   }, []);
 
+  const handleCacheDirectoryChanged = useCallback(async () => {
+    const directory = await cacheAPI.getCacheDirectory();
+    setCacheDirectory(directory);
+    setCacheFolderPref(directory);
+  }, [setCacheFolderPref]);
+
+  const handleMaxCacheSizeChanged = useCallback(async () => {
+    const size = await cacheAPI.getMaxCacheSize();
+    setMaxCacheSize(size);
+    setMaxCacheSizePref(size);
+  }, [setMaxCacheSizePref]);
+
   useEffect(() => {
-    const unbindCacheDirectoryChanged = cacheAPI.subscribeToDirectoryChange(updateCacheDirectory);
-    const unbindMaxCacheSizeChanged = cacheAPI.subscribeToMaxSizeChange(updateMaxCacheSize);
+    const unbindCacheDirectoryChanged = cacheAPI.subscribeToDirectoryChange(handleCacheDirectoryChanged);
+    const unbindMaxCacheSizeChanged = cacheAPI.subscribeToMaxSizeChange(handleMaxCacheSizeChanged);
     const unbindSizeChanged = cacheAPI.subscribeToSizeChange(updateCacheSize);
 
     updateCacheSize();
@@ -44,7 +65,13 @@ export default function CacheProvider(props: CacheProviderProps) {
       unbindMaxCacheSizeChanged();
       unbindSizeChanged();
     };
-  }, [updateCacheSize, updateCacheDirectory, updateMaxCacheSize]);
+  }, [
+    updateCacheSize,
+    updateCacheDirectory,
+    updateMaxCacheSize,
+    handleCacheDirectoryChanged,
+    handleMaxCacheSizeChanged,
+  ]);
 
   const context = useMemo(() => {
     const { getCacheDirectory, getCacheSize, getMaxCacheSize, ...rest } = cacheAPI;
--- a/packages/gui/src/components/nfts/NFTCard.tsx
+++ b/packages/gui/src/components/nfts/NFTCard.tsx
@@ -77,7 +77,10 @@ function NFTCard(props: NFTCardProps) {
               sx={{ zIndex: 1, position: 'absolute', right: 2, top: 2 }}
             />
           )}
-          <NFTPreview id={nftId} disableInteractions={isOffer} ratio={ratio} preview />
+          {/* in selection mode the whole tile selects on click, so media
+              interactions (and the video loop button, which would cover the
+              selection checkbox) are disabled */}
+          <NFTPreview id={nftId} disableInteractions={isOffer || !!onSelect} ratio={ratio} preview />
         </CardActionArea>
         <CardActionArea onClick={() => canExpandDetails && handleClick()} component="div">
           <CardContent>
--- a/packages/gui/src/components/nfts/NFTHashStatus.tsx
+++ b/packages/gui/src/components/nfts/NFTHashStatus.tsx
@@ -31,6 +31,7 @@ export default function NFTHashStatus(props: NFTHashStatusProps) {
   const isLoading = isLoadingNFTVerifyHash || isLoadingNFT;
   const isVerified = preview ? nftPreview?.isVerified : data?.isVerified;
   const error = (errorNFT ?? preview) ? nftPreview?.error : data?.error;
+  const failedFetch = preview ? nftPreview?.failedFetch : data?.failedFetch;
 
   const isValidURI = useMemo(() => {
     if (!nftPreview || !('originalUri' in nftPreview)) {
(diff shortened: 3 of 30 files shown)


emlowe:
@cursor review

hoffmang9:
@bugbot run

wjblanke:
Could you split this into smaller PRs please? "2. Fix unclickable media controls in the NFT detail view" will certainly get in immediately if so and likely the others will be easier to review separately.

jlobue10:
That can certainly be done if that's what Chia Network wants. These are all related though. Cache and NFT loading has issues, video files aren't even playable at all in many instances, and other various quality of life improvements. I am daily driving this version of the modified GUI on CachyOS and after latest commits yesterday evening, things work well for me. Potentially rebasing and simplifying and documenting things better is also on the table. The majority of the code edits were done with Claude Fable, and then ChatGPT Sol found more stuff in an audit. All manually tested and working well on my end.

danieljperry:
> That can certainly be done if that's what Chia Network wants.
Much appreciated, it will make the review process significantly easier.

jlobue10:
> > That can certainly be done if that's what Chia Network wants.
> > Much appreciated, it will make the review process significantly easier.

Alright. Let me start working on the rebase on splitting into smaller PRs. I suggest that the video related ones go together, cache related ones go together and any other UI quality of life ones not directly part of those other two can be split into a third. Does this sound reasonable? The GUI is in great shape for me with these changes.

danieljperry:
> > > That can certainly be done if that's what Chia Network wants.
> > > Much appreciated, it will make the review process significantly easier.
> 
> Alright. Let me start working on the rebase on splitting into smaller PRs. I suggest that the video related ones go together, cache related ones go together and any other UI quality of life ones not directly part of those other two can be split into a third. Does this sound reasonable? The GUI is in great shape for me with these changes.

That sounds great, thanks!

jlobue10:
> > > > That can certainly be done if that's what Chia Network wants.
> > > > Much appreciated, it will make the review process significantly easier.
> > 
> > 
> > Alright. Let me start working on the rebase on splitting into smaller PRs. I suggest that the video related ones go together, cache related ones go together and any other UI quality of life ones not directly part of those other two can be split into a third. Does this sound reasonable? The GUI is in great shape for me with these changes.
> 
> That sounds great, thanks!

Working on it now.

jlobue10:
Closing this PR in favor of the three-way split requested above. All work from this branch now lives in:

1. #3008 — NFT media pipeline 1/3: cache & download infrastructure
2. #3011 — NFT media pipeline 2/3: video playback controls, looping & gallery UX (stacked on #3008; replaces #3009)
3. #3010 — NFT media pipeline 3/3: preview verification & hardening (stacked on #3011)

The three PRs are stacked and meant to be applied together, in order; their combined result is byte-identical to this branch rebased onto current main. Nothing from this PR was dropped. Thanks @danieljperry for the review guidance!

jlobue10:
Closing this PR in favor of the three-way split requested above. All work from this branch now lives in:

1. #3008 — NFT media pipeline 1/3: cache & download infrastructure
2. #3009 — NFT media pipeline 2/3: video playback controls, looping & gallery UX (stacked on #3008)
3. #3010 — NFT media pipeline 3/3: preview verification & hardening (stacked on #3009)

The three PRs are stacked and meant to be applied together, in order; their combined result is byte-identical to this branch rebased onto current main. Nothing from this PR was dropped. Thanks @danieljperry for the review guidance!

PR #3033, opened 2026-08-21
[CHIA-4324] NFT media pipeline 5: download timeout recovery, bounded confirmation previews, buffer cleanup

jlobue10:
Part **5** of the NFT media pipeline series (#3011 → #3010 → #3029). **This PR is meant to be applied after #3029**: it is stacked on that branch, so until #3029 merges its diff includes #3029's 7 commits — **review only the 3 commits listed below.** Each is independent of the others; together they harden the download path that the earlier PRs put in place.

### 1. Retry timed-out NFT downloads once per session (`46ca7ecb9`)

Settling download timeouts as permanent cache errors (#3010) fixed the retry-a-stalled-host-on-every-access loop, but overshot: the timeout message is not in the transient-error list, so one slow first byte from a cold gateway — or a briefly offline machine — wrote an error to the `-info` sidecar that survived restarts. The NFT showed "Preview is not available" forever, until the user cleared the entire NFT cache.

`CacheManager` now keeps an in-memory `timedOutUrls` set: a persisted timeout is retried once per app session. Within the session the URL settles after its first timeout (a stalled host still cannot hold a download slot on every gallery visit), but a restart gets a fresh attempt. `downloadFile` exports `isDownloadTimeoutError`, which matches both timeout messages it produces (inactivity and overall deadline) by the same prefixes earlier sessions persisted — so existing poisoned caches recover on their own.

### 2. Bound WalletConnect confirmation-preview resolution with one overall deadline (`f68778eb5`)

`resolveNftPreviewUrl` walks three ordered URI lists per NFT (image data URIs, metadata URIs, extensionless data URIs) with a 10 s timeout per fetch and no overall limit. URI lists are on-chain data that `nft_add_uri` can extend, and the WalletConnect confirmation dialog only opens after parsing settles — so a `take_offer` referencing NFTs with many dead or slow hosts kept the security dialog off screen for 10 s × URIs × lists, minutes in the worst case, while the request could expire.

All fallbacks for one NFT now share a 20 s budget: `nftGetMetadata` / `nftGetImageDataUrl` take a `timeoutBudget`, each fetch gets the remaining time (capped at its own per-fetch timeout), and the walk stops once the budget is spent. Previews degrade to the placeholder in that case — the dialog itself is never held up by more than the budget. NFTs resolve concurrently, so this bounds the whole parse regardless of how many NFTs an offer references.

### 3. Drop redundant buffer copies and a dead hook export (`72a9b8d58`)

The verified-preview path copied every downloaded byte twice for no gain: `fetchBuffer` re-allocated each response chunk element-by-element via `Uint8Array.from` before collecting it (Electron's `net` module delivers fresh Buffers, so retaining them is safe and `Buffer.concat` already performs the single final copy), and the metadata checksum round-tripped the whole buffer through a `latin1` string before hashing — an up-to-10 MB string allocation to produce the digest that hashing the Buffer directly yields.

Also removes the unused `useNFTVideoLoop` default export: both consumers import the named hooks and derive the effective global-or-per-video state themselves, so the wrapper only added surface that could drift from the real logic.

No user-facing strings or settings are added; no cache migration is needed (cache keys and sidecar formats are unchanged).

## Testing

- New `CacheManager` test: a timeout persisted by one session (`Request timed out after …ms of inactivity` in the `-info` sidecar) is retried, and succeeds, in a fresh `CacheManager` instance against the same cache directory.
- New `parseCommandDisplay` test: once a slow metadata host consumes the whole budget, the remaining metadata fallbacks are never tried and the walk resolves to the placeholder; the existing resolution tests now also assert that a numeric budget is passed through to every `nftGetMetadata` / `nftGetImageDataUrl` call.
- Full gui jest suite green (355 tests); eslint/prettier clean on all touched files.

https://claude.ai/code/session_01YKEGodEgVdEuvUM8dza52q

---

> [!NOTE]
> **Medium Risk**
> Touches NFT fetch/cache behavior and WalletConnect confirmation previews, including an opt-in third-party gateway. Content remains hash-verified and the gateway is off by default, but cache retry and dialog timing changes are easy to get wrong.
> 
> **Overview**
> Adds an **opt-in** NFT setting to fetch `ipfs://` media through the public `ipfs.io` HTTPS gateway (off by default). Content is still hash-checked against the on-chain URI; Electron cannot request the `ipfs:` scheme directly.
> 
> Network call sites (`downloadFile`, `fetchBuffer`, `fetchJSON`, Chromium downloads) go through `toFetchableUrl`. With the option off, IPFS fetches throw `IpfsGatewayDisabledError` and are **not** persisted as cache errors, so turning the option on retries cleanly. Cached bytes stay keyed by the original URI. Toggling the setting retries failed in-memory metadata and re-runs hash verification.
> 
> Also hardens the download/preview path: persisted download timeouts are retried **once per app session**; WalletConnect confirmation previews share a **20s** fallback budget so slow URI lists cannot delay the security dialog; `fetchBuffer`/checksum drop extra copies; unused `useNFTVideoLoop` default export is removed.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 72a9b8d580bf6ef5102ab727a8de6db753fb0fd9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/gui/src/components/nfts/NFTHashStatus.tsx
+++ b/packages/gui/src/components/nfts/NFTHashStatus.tsx
@@ -37,19 +37,18 @@ export default function NFTHashStatus(props: NFTHashStatusProps) {
   const failedFetch = preview ? nftPreview?.failedFetch : data?.failedFetch;
 
   const isValidURI = useMemo(() => {
-    if (!nftPreview || !('originalUri' in nftPreview)) {
+    const uri = nftPreview?.uri;
+    if (!uri) {
+      // nothing to validate — other branches cover the missing-preview cases
       return true;
     }
 
-    if (nftPreview.uri) {
-      // While the user has IPFS gateway fetching enabled, ipfs:// URIs are
-      // served through an HTTPS gateway by the cache layer, so validate the
-      // gateway form instead of flagging them as invalid. With the option
-      // off they are not fetchable and stay flagged.
-      return isValidURL(ipfsGateway ? ipfsToGatewayUrl(nftPreview.uri) : nftPreview.uri);
-    }
-
-    return false;
+    // While the user has IPFS gateway fetching enabled, ipfs:// URIs are
+    // served through an HTTPS gateway by the cache layer, so validate the
+    // gateway form instead of flagging them as invalid. With the option off
+    // they are not fetchable and stay flagged — unless the file already
+    // verified from the cache, which the message branches above this check.
+    return isValidURL(ipfsGateway ? ipfsToGatewayUrl(uri) : uri);
   }, [nftPreview, ipfsGateway]);
 
   const icon = useMemo(() => {
--- a/packages/gui/src/components/nfts/NFTHashStatus.tsx
+++ b/packages/gui/src/components/nfts/NFTHashStatus.tsx
@@ -6,6 +6,7 @@ import { Chip, Typography } from '@mui/material';
 import CircularProgress from '@mui/material/CircularProgress';
 import React, { useMemo } from 'react';
 
+import useIpfsGateway from '../../hooks/useIpfsGateway';
 import useNFT from '../../hooks/useNFT';
 import useNFTVerifyHash from '../../hooks/useNFTVerifyHash';
 import ipfsToGatewayUrl from '../../util/ipfs';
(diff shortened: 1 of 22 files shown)


danieljperry:
@cursor review

emlowe (approved):

